PR #2, opened 2020-02-05
Basic package setup

henryiii:
Changes:
* Add CONTRIBUTING.md, with environment setup instructions
* Add basic GitHub Actions tests
* Restructure to best practices for Python package
* Use config mode for setuptools
* Use PEP 517 mode (optionally)
* Add pre-commit and black
* Set up basic test infrastructure
* Add a ROOT enabled conda env, for ROOT comparison tests
* Added ROOT testing environment


environment.yml
```diff
@@ -0,0 +1,10 @@
+name: vector
+channels:
+  - conda-forge
+dependencies:
+  - jupyterlab
+  - pip
+  - pytest
+  - root >=6.18.04
```
eduardo-rodrigues: I guess you're adding ROOT here for tests rather than for the normal user. I think that's fine, at least for now.

henryiii: A normal user should do pip install vector. And for the eventual binder examples, there will probably be a comparison with ROOT.

PR #4, opened 2020-03-04
Working on Awkward-ready Vectors

henryiii:
First adaptation of work from @jpivarski


src/vector/awkward.py
```diff
@@ -0,0 +1,42 @@
...
+
+class LorentzXYZ(ak.Record, LorentzXYZCommon):
+    def __repr__(self):
+        return "Lxyz({0:.3g} {1:.3g} {2:.3g} {3:.3g})".format(self.x, self.y, self.z, self.t)
+
+class LorentzXYZArray(ak.Array, LorentzXYZCommon):
+    pass
+
+
+# This function only works as a ufunc overload, but it
```
eduardo-rodrigues: Seems some words got deleted

src/vector/core/lorentz/xyzt.py
```diff
@@ -0,0 +1,19 @@
...
+# Distributed under the 3-clause BSD license, see accompanying file LICENSE
+# or https://github.com/scikit-hep/vector for details.
+
+
+# Functions that return a value
+
+def pt(rec):
+    return np.sqrt(rec.x**2 + rec.y**2)
+
+def eta(rec):
```
eduardo-rodrigues: One can be even more concise here and use `pt` in the denominator.

BTW, do we want to test for `pt==0` and return `inf`? Seems appropriate unless numpy does that for you automatically when calculating the ratii array.

eduardo-rodrigues:
This is looking good!

src/vector/core/lorentz/xyzt.py
henryiii: It does:

```python
In [2]: np.array([1]) / np.array([0])
/usr/local/Cellar/ipython/7.12.0/libexec/bin/ipython:1: RuntimeWarning: divide by zero encountered in true_divide
  #!/usr/local/opt/python/bin/python3.7
Out[2]: array([inf])
```

src/vector/awkward.py
henryiii: That was missing in Jim's original example as well.

src/vector/core/lorentz/xyzt.py
eduardo-rodrigues: OK, great - had not checked.

src/vector/awkward.py
eduardo-rodrigues: Not a reason not to improve, right? At least let's remove the now-spurious ", but it".

henryiii:
Putting this in as a good starting point.

PR #5, opened 2020-03-09
Bump Scikit-hep-testdata

henryiii:
Should remove a warning and work a bit better.


henryiii:
GHA seems to be down. Will retry later.

Issue #8, opened 2020-03-26
pip failed in environment.yml


LovelyBuggies:
I'm trying to build the conda dev env.

```bash
conda env create
jupyter lab
conda activate vector
conda config --env --add channels conda-forge
```

But `ModuleNotFoundError: No module named 'awkward1'` in notebooks/VectorDesignPrototype.ipynb.

P.S., I am using Python3 kernel. Is this related to jupyter kernel?

henryiii:
You will need  `pip install awkward1` for now. Once awkward1 is finalized, we can provide a conda package for it; it's *just* finishing being developed.

LovelyBuggies:
I think the `conda activate` has already installed the package. I run the following **after** the above commands:

```python3
(vector) NinodeMacBook-Pro:vector nino$ pip install awkward1
Requirement already satisfied: awkward1 in /Users/nino/anaconda3/envs/vector/lib/python3.8/site-packages (0.2.7)
Requirement already satisfied: numpy>=1.13.1 in /Users/nino/anaconda3/envs/vector/lib/python3.8/site-packages (from awkward1) (1.18.1)
```

And the notebooks still cannot import awkward1 for some reason.

LovelyBuggies:
I am confused about the `enviroment.yml` when activating. I see it says:

```yml
  - pip:
    - "awkward1>=0.2.3,!=0.2.4"
    - "scikit-hep-testdata>=0.2.0"
    - -e .
```

Why do I not get `awkward1`, or why do I get `awkward1` 0.2.7 as shown (it should be 0.2.3 and not 0.2.4, shouldn't it)?

jpivarski:
If `awkward1>=0.2.3,!=0.2.4` was written with the intention of `awkward1==0.2.3`, then it no longer works because there are higher versions than `0.2.4` now. I'm not releasing them as often as I did, but I just released `0.2.10` today.

What's holding you back to `0.2.3`? It's probably something that's been fixed. But the most immediate answer is that you probably want `awkward1==0.2.3`.

LovelyBuggies:
Yep, I think this has something to do with jupyter kernel, I manually created a kernel named `vector` and used this. Then the notebooks worked.

LovelyBuggies:
@jpivarski , why should use 0.2.3?

> What's holding you back to 0.2.3? It's probably something that's been fixed. But the most immediate answer is that you probably want awkward1==0.2.3.

[...]

PR #9, opened 2020-03-27
dev env for notebooks

LovelyBuggies:


LovelyBuggies:
Development environment for notebooks #8

LovelyBuggies:
> Whatever motivated the restriction to 0.2.3 is probably fixed by now or would be an easy fix, so you should switch to the latest version at your convenience.

Mentioned in #8 .

LovelyBuggies:
@henryiii LGTM, anything else to change?

.github/CONTRIBUTING.md
```diff
@@ -4,9 +4,10 @@
 You should *always* use a virtual environment when developing software. Setup:
 
 ```bash
-python3 -m venv .env
+python -m venv .env
```
henryiii: Why are you dropping the 3? According to [PEP 0394](https://www.python.org/dev/peps/pep-0394/), `python` is not supposed to point at `python3` *still* (except in a virtual env), and you have to use Python 3 to use venv. (The dual-Python package you can install that does venv's is called `virtualenv` instead of `venv`.)

Modern systems are looking at just dropping `python` and providing `python2` and `python3` (such as RHEL 8).

environment.yml
```diff
@@ -2,14 +2,14 @@ name: vector
 channels:
   - conda-forge
 dependencies:
-  - jupyterlab >=1.2
```
henryiii: I believe this spacing is the Conda recommended spacing. It's used in all the Conda-forge recipe's, anyway.

environment.yml
```diff
@@ -2,14 +2,14 @@ name: vector
...
   - pytest
-  - root >=6.18.04
-  - uproot >=3.11
-  - numba >=0.46
+  - root>=6.18.04
+  - uproot>=3.11
+  - numba>=0.46
   - pip:
-    - "awkward1>=0.2.3,!=0.2.4"
+    - "awkward1>=0.2.3"
```
henryiii: Let's just make this 0.2.5 - there was a bad bug in 0.2.4 that broke vector, which is why it was removed here. Yes, we can use 0.2.3 currently, but there's no reason to include it.

.github/CONTRIBUTING.md
LovelyBuggies: Actually, "Python3 -> Python" is recommended by Jim in https://github.com/scikit-hep/aghast/pull/31#discussion_r399652299. If you are sure using `python3` in command lines is a better manner. I will inform Jim to make some changes to Aghast, too.

jpivarski: In Aghast, I asked Nino to drop the "3" in "python3" because that's particular to some OS distributions. I don't know how many systems do it one way and how many do it the other, but I would have thought it would be more common nowadays for plain old "python" to actually be Python 3. (It is on mine, for instance.)

These are instructions for a developer. I think one can just say, "Make sure to use Python 3" and however they do that on their system, they should do that.

henryiii: All Python 3 instances will always have `python3`. It is invalid according to PEP 0394 to have `python` be 3, but some systems have an "alternatives" system to change this, and a few modern systems just do this anyway, PEP or no PEP. In this case, we are running a Python 3 only builtin module (venv), not running setup.py.

In general, use python3 if the command only works on Python 3, python2 if it only works on 
Python 2, and python if it works in either.

If this was not aimed at developers, I'd probably point out there's a better but not builtin package called virtualenv.

henryiii: See https://linuxize.com/post/how-to-install-python-on-centos-8/

henryiii: I just realized the PEP was updated in 2019 to remove the requirements on `python` pointing at python2, and basically say anything goes. Good. Recommendation for us is still the same though. Jim, are you on a non LTS Ubuntu, perhaps, from 2019? Will be interesting if 20.04 has a Python 3 `python` command.

LovelyBuggies: Python2 does not have venv. It has virtualenv.

So I think this should be `python -m venv .env`.

LovelyBuggies:
@henryiii The CI failed again. It's odd, I saw all checks had passed before it was merged.

henryiii:
The conda job that tests ROOT sometimes times out (6 hours!) instead of finishing in a few minutes. Not sure why yet. We can replace it with a docker powered job if it continues to fail.

chrisburr:
It looks like it fails when installing miniconda itself? If so that isn't necessary as conda is already installed in the GitHub actions VMs. The only think you might need is to source the shell integration:

```bash
eval "$(conda shell.bash hook)"
```

henryiii:
Okay, so step 1 is to try Conda directly, and step 2 is try a docker version.

LovelyBuggies:
@henryiii I came back to see this PR again and found that PR #11 could not help.

PR #19, opened 2020-12-09
awkward1 to awkward

kreczko:
Since awkward 1.0.0 has been released, changing `awkward1` &rarr; `awkward`.
Without these changes any code using both awkward1 (e.g. from `vector`) and awkward 1.0 will crash.

Also updated the pytest fixture for the new syntax since the last awkward1 change.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
         python-version:
         - 2.7
         - 3.6
-        - 3.9
+        - 3.8
     name: Check Python ${{ matrix.python-version }}
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,17 +9,6 @@ on:
     - develop
 
 jobs:
-  pre-commit:
-    runs-on: ubuntu-latest
-    steps:
-    - uses: actions/checkout@v2
-      with:
-        fetch-depth: 0
-    - uses: actions/setup-python@v2
-    - uses: pre-commit/action@v2.0.0
-      with:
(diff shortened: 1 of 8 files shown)


kreczko:
hmm, I am still seeing
```
  /software/miniconda/envs/hep/lib/python3.7/site-packages/numba/core/dispatcher.py:237: UserWarning: Numba extension module 'awkward1._connect._numba' failed to load due to 'ImportError(generic_type: type "kernel_lib" is already registered!)'.
    entrypoints.init_all()
```
but I cannot figure out where this is coming from

henryiii:
Oh, that's nice, that's the first time I've seen pre-commit.ci do the auto-fix. :)

Thanks for fixing this; I'll have a PR later today that will fill out a bit more functionality, too.

henryiii:
@jpivarski - the awkward1 numba extension might be a left over from the name change that didn't get uninstalled when upgrading?

kreczko:
ah, yes; good point @henryiii. Uninstalled `awkward1` - no more issues

jpivarski:
Numba extensions are defined as "`entry_points`", which affect Numba at `import numba` time, regardless of whether the package that defined the entry points has been imported. I guess it's a problem if both `awkward1` and `awkward>=1.0.0` define the same entry points. Either removing the old `awkward1` or upgrading `awkward1` to its latest version (which does not define entry points) should solve the issue.

henryiii:
You forgot to bump uproot to being pinned to 3 to 4+, apparently awkward 1.0 requires uproot 4 while awkward1 did not. It causes the pip "will break in 20.3" warning on pip 20.2. I've fixed it in #20.

Thanks for your help, it was much appreciated!!!

PR #31, opened 2021-03-03
Removed the code, keeping most of the project structure.

jpivarski:


jpivarski:
This won't work until I put some tests back in. I won't merge the PR until that happens.

jpivarski:
This is a good first step. It implements what I tried to [describe in the wiki](https://github.com/scikit-hep/vector/wiki).

PR #33, opened 2021-03-05
Fill out Object and NumPy implementations.

jpivarski:
Don't laugh:


jpivarski:
The above diagram has been implemented for *Object and *Numpy backends, with the following properties:

   * x
   * y
   * rho
   * phi
   * rho2
   * z
   * theta
   * eta
   * costheta
   * mag
   * mag2
   * t
   * t2
   * tau
   * tau2

PR #51, opened 2021-03-30
Rename 'methods', 'backends', 'compute' to start with underscores.

jpivarski:
@henryiii I looked at what users should be expected to access, and concluded that only the symbols exposed as `vector.*` are public. Some of these symbols come from `vector.methods` and `vector.backends.*` and none of them come from `vector.compute.*`. Since these modules shouldn't be accessed directly, I guess they should start with underscores. This PR only renames them.

Since you're working on another PR, it might need to be carefully merged with this (or you might have opinions about this drastic attempt to hide almost everything).


henryiii:
> or you might have opinions about this drastic attempt to hide almost everything

No problem, I'm strongly for hiding anything that is not expected to be public.

> might need to be carefully merged with this

No problem. Though make sure you have two commits, one with the move, one with the rename. Never edit a file and move it at the same time in Git, the tracking algorithm struggles if you do that. If you need me to, I can resquash things into two commits before a rebase-and-merge.

jpivarski:
> If you need me to, I can resquash things into two commits before a rebase-and-merge.

That would be helpful, thanks. (I've never _successfully_ done a rebase.)

jpivarski:
Nevermind: I'll merge #52 and then do a new rename with move and change as separate commits in a new PR.

PR #59, opened 2021-03-31
docs: prepare

henryiii:
Working on docs.

- docs: prepare for docs
- docs: work on setting up docs


henryiii:
This is not fixed yet, but better than before. Probably better to get this in, then iterate. There still are a few hundred warnings and such.

docs/conf.py
```diff
@@ -48,23 +57,36 @@
...
 
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
 #
-html_theme = "sphinx_rtd_theme"
+html_theme = "sphinx_book_theme"
+
+# Config for the Sphinx book
+
+html_baseurl = "https://boost-histogram.readthedocs.io/en/latest/"
```
henryiii, line 71:
```suggestion
html_baseurl = "https://vector.readthedocs.io/en/latest/"
```

docs/conf.py
```diff
@@ -48,23 +57,36 @@
...
+
+
+html_theme_options = {
+    "home_page_in_toc": True,
+    "repository_url": "https://github.com/scikit-hep/boost-histogram",
```
henryiii, line 76:
```suggestion
    "repository_url": "https://github.com/scikit-hep/vector",
```

You can seem my starting point. :)

PR #63, opened 2021-04-01
fix(types): use static inheritance instead of dynamic

henryiii:
Goodbye, 78 `type: ignores`. :)

### DIFF
--- a/src/vector/_backends/object_.py
+++ b/src/vector/_backends/object_.py
@@ -55,53 +55,53 @@ class TemporalObject(CoordinatesObject, Temporal):
     pass
 
 
-class AzimuthalObjectXY(typing.NamedTuple):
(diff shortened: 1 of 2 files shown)


src/vector/_backends/object_.py
```diff
@@ -55,85 +55,82 @@ class TemporalObject(CoordinatesObject, Temporal):
...
     y: float
 
+
+class AzimuthalObjectXY(AzimuthalObject, AzimuthalXY, TupleXY):
     @property
     def elements(self) -> typing.Tuple[float, float]:
         return (self.x, self.y)
 
 
-AzimuthalObjectXY.__bases__ = (AzimuthalObject, AzimuthalXY, tuple)
```
henryiii, line 67: This is the key change.

PR #64, opened 2021-04-01
chore(types): allow mypy to verify dispatch exists

henryiii:
This drops several hundred more type ignores; without NumPy fixes, probably can't drop many more than this. 64 type ignores left. Even the stdlib has some, so the goal is not to get all the way to 0.

MyPy can see the dispatch function in the `_compute` modules now, and fixed a couple of issues with types and simplified the way ignoring missing type signatures works in those internal modules.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -84,6 +84,10 @@ warn_return_any = False
 no_implicit_reexport = True
 strict_equality = True
 
+[mypy-vector._compute.*.*]
+disallow_untyped_defs = False
+disallow_untyped_calls = False
+
 [mypy-numba.*]
 ignore_missing_imports = True
 
--- a/src/vector/_methods.py
+++ b/src/vector/_methods.py
@@ -1222,7 +1222,7 @@ class Vector(VectorProtocol):
 
         return self._wrap_result(
             type(self),
-            (planar.x.dispatch(self), planar.y.dispatch(self)),  # type: ignore
+            (planar.x.dispatch(self), planar.y.dispatch(self)),
             [AzimuthalXY, None],
             1,
         )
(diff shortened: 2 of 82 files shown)


src/vector/_methods.py
```diff
@@ -1222,7 +1222,7 @@ def to_xy(self) -> VectorProtocolPlanar:
 
         return self._wrap_result(
             type(self),
-            (planar.x.dispatch(self), planar.y.dispatch(self)),  # type: ignore
+            (planar.x.dispatch(self), planar.y.dispatch(self)),
```
henryiii, line 1225: These are the removed type ignores.

src/vector/_methods.py
```diff
@@ -2408,8 +2408,8 @@ def _lib_of(*objects: VectorProtocol) -> Module:  # NumPy-like module
 def _from_signature(
     name: str,
     dispatch_map: typing.Dict[typing.Any, typing.Any],
-    signature: typing.Tuple[typing.Any],
-) -> typing.Tuple[typing.Any]:
+    signature: typing.Tuple[typing.Any, ...],
+) -> typing.Tuple[typing.Any, ...]:
```
henryiii, line 2412: And this is the signature fix; `Tuple[Any, ...]` is a generic tuple. `Tuple[Any]` is a one-element tuple with anything in it.

setup.cfg
```diff
@@ -84,6 +84,10 @@ warn_return_any = False
 no_implicit_reexport = True
 strict_equality = True
 
+[mypy-vector._compute.*.*]
+disallow_untyped_defs = False
+disallow_untyped_calls = False
+
```
henryiii, line 90: Untyped defs and calls are allowed in these modules.

PR #76, opened 2021-04-09
fix: cannot import without Awkward installed

### LINKED ISSUE #75: `awkward` required but not installed on `pip install vector`
congrats on the release! 

I see the following on a fresh env - is this intended?
```
python -m venv _testenv
source _testenv/bin/activate
pip install vector
python -m vector
```

produces

```
Traceback (most recent call last):
  File "/Users/lukasheinrich/.pyenv/versions/3.7.2/lib/python3.7/runpy.py", line 183, in _run_module_as_main
    mod_name, mod_spec, code = _get_module_details(mod_name, _Error)
  File "/Users/lukasheinrich/.pyenv/versions/3.7.2/lib/python3.7/runpy.py", line 142, in _get_module_details
    return _get_module_details(pkg_main_name, error)
  File "/Users/lukasheinrich/.pyenv/versions/3.7.2/lib/python3.7/runpy.py", line 109, in _get_module_details
    __import__(pkg_name)
  File "/Users/lukasheinrich/Code/pylhe/_testenv/lib/python3.7/site-packages/vector/__init__.py", line 8, in <module>
    from vector._backends.awkward_ import VectorAwkward
  File "/Users/lukasheinrich/Code/pylhe/_testenv/lib/python3.7/site-packages/vector/_backends/awkward_.py", line 35, in <module>
    import awkward as ak
ModuleNotFoundError: No module named 'awkward'
```

henryiii:
Closes #75. There are other possible solutions, @jpivarski, if you prefer something different.

### DIFF
--- a/src/vector/__init__.py
+++ b/src/vector/__init__.py
@@ -5,7 +5,6 @@
 
 import typing
 
-from vector._backends.awkward_ import VectorAwkward
 from vector._backends.numpy_ import (
     MomentumNumpy2D,
     MomentumNumpy3D,
@@ -50,7 +49,16 @@ from vector._methods import (
 )
 from vector.version import version as __version__
 
-__all__ = (
+try:
+    import awkward
+except ModuleNotFoundError:
+    awkward = None
+    if not typing.TYPE_CHECKING:
+        VectorAwkward = None
+else:
+    from vector._backends.awkward_ import VectorAwkward
+
+__all__: typing.Tuple[str, ...] = (
     "Array",
     "Azimuthal",
     "AzimuthalRhoPhi",
@@ -98,7 +106,11 @@ __all__ = (
 
 
 def __dir__() -> typing.Tuple[str, ...]:
-    return __all__
+    return (
+        tuple(s for s in __all__ if s != "VectorAwkward")
+        if awkward is None
+        else __all__
+    )
 
 
 def register_numba() -> None:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,7 @@ jobs:
     name: Check Python ${{ matrix.python-version }}
     steps:
     - uses: actions/checkout@v2
+
     - uses: actions/setup-python@v2
       with:
         python-version: ${{ matrix.python-version }}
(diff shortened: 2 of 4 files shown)


src/vector/__init__.py
```diff
@@ -5,7 +5,6 @@
 
 import typing
 
-from vector._backends.awkward_ import VectorAwkward
```
jpivarski, line 8: This line was a last-minute mistake: it was designed up to this point to not import `vector._backends.awkward_` at startup, only `vector._backends.numpy_`. I must have forgotten that in some later step.

Would it be preferable to have `VectorAwkward` be undefined, rather than `None`, if `awkward` can't be imported? That's more in line with what I'd expect.

Independently of that, it may be better to do the test for importability like this:

```python
try:
    import awkward
except ModuleNotFoundError:
    pass   # or VectorAwkward = None
else:
    from vector._backends.awkward_ import VectorAwkward
```

That way, difficulties importing `awkward` are distinguishable from difficulties importing `vector._backends.numpy_`, following the "do as little as possible in a `try` block" rule. Specifically, it's quite normal for `awkward` to not be installed on a given system, but assuming that it is installed, being unable to import one of Vector's own submodules would be some kind of bug.

jpivarski (commented):
Sorry I missed this earlier. I have a question (should `VectorAwkward` be `None` or undefined?) and a suggestion (distinguish failure to import `awkward` from potential failures to import Vector's submodule).

src/vector/__init__.py
henryiii: The problem with not having defined at all is `__all__`, which probably should not be dynamic (we certainly would have to trick MyPy if it was dynamic, as mypy needs to know if `vector.VectorAwkward` exists without knowing if awkward is installed; flake8 likely would be unhappy too, etc). I can experiment with making it dynamic and see what happens. Removing it from `__all__` at runtime would make `__dir__ `work better, so maybe that is better.

This is at the end, so it _should_ be pretty safe, but I still like your version above better.

henryiii: MyPy is happy, but flake8 blows up.

henryiii: Implemented an intermediate solution. It's still set to `None`, but doesn't show up in `dir()` / IPython (Python 3.7+ only). Also added "simple" tests to make sure it works without Awkward / Numba installed.

PR #82, opened 2021-04-12
fix: check awkward version

### LINKED ISSUE #81: Add version check for awkward
We should add a version check for awkward array, based on packaging + importlib_metadata/importlib.metadata. Also add an awkward extra.

_Originally posted by @jpivarski in https://github.com/scikit-hep/vector/issues/79#issuecomment-818031030_

henryiii:
Closes #81.

Had a little trouble installing Awkward to test it (https://github.com/scikit-hep/awkward-1.0/issues/821), but seems to be working. Edit: fixed.

### DIFF
--- a/src/vector/__init__.py
+++ b/src/vector/__init__.py
@@ -58,17 +58,23 @@ else:
     import importlib.metadata as importlib_metadata
 
 
-try:
+def _import_awkward():
     import awkward
-except ModuleNotFoundError:
-    awkward = None
+    awk_version = packaging.version.Version(importlib_metadata.version("awkward"))
+    if awk_version < packaging.version.Version("1.2.0rc5"):
+        msg = f"""awkward {awk_version} is too old; please upgrade to 1.2.0 or later:
+
+    pip install -U awkward
+"""  # the only context users will see this message is if they're trying to use vector.awk
+        raise ImportError(msg)
+    return awkward
+
+try:
+    _import_awkward()
+except ImportError:  # superclass of ModuleNotFoundError, I believe
     if not typing.TYPE_CHECKING:
         VectorAwkward = None
 else:
-    awk_version = packaging.version.Version(importlib_metadata.version("awkward"))
-    if awk_version < packaging.version.Version("1.2.0rc5"):
-        msg = f"Awkward {awk_version} installed and too old, remove or install 1.2.0+"
-        raise ImportError(msg)
     from vector._backends.awkward_ import VectorAwkward
 
 
--- a/src/vector/__init__.py
+++ b/src/vector/__init__.py
@@ -58,20 +58,21 @@ else:
     import importlib.metadata as importlib_metadata
 
 
-def _import_awkward():
-    import awkward
+def _import_awkward() -> None:
     awk_version = packaging.version.Version(importlib_metadata.version("awkward"))
     if awk_version < packaging.version.Version("1.2.0rc5"):
-        msg = f"""awkward {awk_version} is too old; please upgrade to 1.2.0 or later:
-
-    pip install -U awkward
-"""  # the only context users will see this message is if they're trying to use vector.awk
+        # the only context users will see this message is if they're trying to use vector.awk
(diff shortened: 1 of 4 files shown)


jpivarski:
Before you merge this, let me make a suggestion...

src/vector/__init__.py
```diff
@@ -49,15 +52,26 @@
...
 except ModuleNotFoundError:
     awkward = None
     if not typing.TYPE_CHECKING:
         VectorAwkward = None
 else:
+    awk_version = packaging.version.Version(importlib_metadata.version("awkward"))
+    if awk_version < packaging.version.Version("1.2.0rc5"):
+        msg = f"Awkward {awk_version} installed and too old, remove or install 1.2.0+"
+        raise ImportError(msg)
     from vector._backends.awkward_ import VectorAwkward
```
jpivarski, line 72:
```suggestion
def _import_awkward():
    import awkward
    awk_version = packaging.version.Version(importlib_metadata.version("awkward"))
    if awk_version < packaging.version.Version("1.2.0rc5"):
        msg = f"""awkward {awk_version} is too old; please upgrade to 1.2.0 or later:

    pip install -U awkward
"""  # the only context users will see this message is if they're trying to use vector.awk
        raise ImportError(msg)
    return awkward

try:
    _import_awkward()
except ImportError:  # superclass of ModuleNotFoundError, I believe
    if not typing.TYPE_CHECKING:
        VectorAwkward = None
else:
    from vector._backends.awkward_ import VectorAwkward
```

and replace

```python
import awkward
```

with

```python
awkward = vector._import_awkward()
```

in `vector/_backends/awkward_.py`.

henryiii: Soon we are hoping to make `VectorAwkward(...)` a way to construct? But we can adjust when that is ready. Also, I don't really like putting the `pip install` command directly in the message, what if a user is using conda? Or pipenv, poetry, etc... Also, `awkward = function()` will mess up static typing and `flake8`.

Would it be that bad to have a `vector.awkward`, and all awkward functionality requires `import vector.awkward`? All the workarounds would go away and the user could decide if they need to pay the cost of importing awkward. (not a patch release candidate, obviously)

henryiii: I'm going to merge this change as is, then modify a bit, to keep a comparison easy to find (we'll squash and merge at then end, anyway)

jpivarski: I've been putting pip and conda install suggestions in Awkward and Uproot ImportErrors: https://github.com/scikit-hep/uproot4/blob/main/src/uproot/extras.py

If someone's using an alternative, they should be aware of that and made the appropriate adjustments (including two, pip and conda, is the first step in communicating that there _are_ alternatives). Beyond the help for novices, who might otherwise start downloading tarballs from GitHub if they're not nudged in the right direction, it's great to be able to triple-click a suggestion and paste it back into the terminal. I never set up PR branches; I `git push` knowing that it will fail because git gives me an error message that includes the right command. It's useful.

It wouldn't be bad at all to have a `vector.awkward` that users have to manually import. Both designs involve an opt-in, though one is just a function call and the other is an import and a function call. It's a complexity cost they'd be seeking to avoid, I think, since the time to `import awkward` is 0.25 seconds, below threshold for humans.

henryiii:
My changes are here: https://github.com/scikit-hep/vector/pull/82/commits/c2c0540fffcd197e83962f0cc90869b5191b92de . Should be good for now.

src/vector/__init__.py
henryiii: `pip install -U awkward` is the wrong thing to do if they have installed with conda, and will really mess their environment/system up. They had to know how to install vector, so they should also apply whatever they know to installing awkward. (Updating adds a tiny bit of complexity, true, but not much. Also, `pip install awkward>=1.2` might be better.).

henryiii: Also, hopefully, long term there won't be many old awkwards around.

PR #89, opened 2021-04-15
feat: add constructors for `VectorObject2D` and `MomentumObject2D`

### LINKED ISSUE #88: feat: new constructor
Placeholder issue for the constructor work, main feature of 0.9.

henryiii:
Closes #88.

WIP, currently just 2D Object implemented.

### DIFF
--- a/src/vector/backends/object.py
+++ b/src/vector/backends/object.py
@@ -666,7 +666,7 @@ class VectorObject2D(VectorObject, Planar, Vector2D):
         return cls(azimuthal=AzimuthalObjectRhoPhi(rho, phi))
 
     def __init__(
-        self, azimuthal: typing.Optional[AzimuthalObject] = None, **kwargs: float
+        self, azimuthal: AzimuthalObject | None = None, **kwargs: float
     ) -> None:
 
         for k, v in kwargs.copy().items():
--- a/src/vector/backends/object.py
+++ b/src/vector/backends/object.py
@@ -60,6 +60,7 @@ from vector._methods import (
     _handler_of,
     _ltype,
     _repr_generic_to_momentum,
+    _repr_momentum_to_generic,
     _ttype,
 )
 from vector._typeutils import FloatArray
@@ -667,6 +668,11 @@ class VectorObject2D(VectorObject, Planar, Vector2D):
     def __init__(
         self, azimuthal: typing.Optional[AzimuthalObject] = None, **kwargs: float
     ) -> None:
+
+        for k, v in kwargs.copy().items():
+            kwargs.pop(k)
+            kwargs[_repr_momentum_to_generic.get(k, k)] = v
+
         if not kwargs and azimuthal is not None:
             self.azimuthal = azimuthal
         elif kwargs and azimuthal is None:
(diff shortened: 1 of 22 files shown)


src/vector/backends/object.py
```diff
@@ -799,17 +830,47 @@ class MomentumObject2D(PlanarMomentum, VectorObject2D):
...
+        elif kwargs and azimuthal is None:
+            if set(kwargs) == {"px", "py"}:
+                self.azimuthal = AzimuthalObjectXY(kwargs["px"], kwargs["py"])
+            elif set(kwargs) == {"pt", "phi"}:
+                self.azimuthal = AzimuthalObjectRhoPhi(kwargs["pt"], kwargs["phi"])
+            else:
+                raise TypeError("invalid arguments, must be px=, py= or pt=, phi=")
+        else:
+            raise TypeError("must give Azimuthal if not giving keyword arguments")
+
```
Saransh-cpp, line 866: I went through the overloads of `vector.obj` and a user can indeed create a `MomentumObject` by specifying a mixture of momentum and generic coordinates. I think this should be removed and the momentum coordinate names must be converted to generic in the `__init__` method of `VectorObject` classes, which will then be inherited by `MomentumObject` classes.

src/vector/backends/object.py
```diff
@@ -662,17 +681,29 @@ def from_rhophi(cls, rho: float, phi: float) -> "VectorObject2D":
...
+            self.azimuthal = azimuthal
+        elif kwargs and azimuthal is None:
+            if set(kwargs) == {"x", "y"}:
+                self.azimuthal = AzimuthalObjectXY(kwargs["x"], kwargs["y"])
+            elif set(kwargs) == {"rho", "phi"}:
+                self.azimuthal = AzimuthalObjectRhoPhi(kwargs["rho"], kwargs["phi"])
+            else:
+                raise TypeError("invalid arguments, must be x=, y= or rho=, phi=")
+        else:
+            raise TypeError("must give Azimuthal if not giving keyword arguments")
```
Saransh-cpp, line 701: Removing the `MomentumObject` constructor and changing the `momentum` coordinates to `generic` ones works!
```suggestion
    def __init__(
        self, azimuthal: typing.Optional[AzimuthalObject] = None, **kwargs: float
    ) -> None:
        for k, v in kwargs.copy().items():
            kwargs.pop(k)
            kwargs[_repr_momentum_to_generic.get(k, k)] = v

        if not kwargs and azimuthal is not None:
            self.azimuthal = azimuthal
        elif kwargs and azimuthal is None:
            if set(kwargs) == {"x", "y"}:
                self.azimuthal = AzimuthalObjectXY(kwargs["x"], kwargs["y"])
            elif set(kwargs) == {"rho", "phi"}:
                self.azimuthal = AzimuthalObjectRhoPhi(kwargs["rho"], kwargs["phi"])
            else:
                raise TypeError("invalid arguments, must be x=, y= or rho=, phi=")
        else:
            raise TypeError("must give Azimuthal if not giving keyword arguments")
```

Saransh-cpp:
I think this PR should work now, given that we are fine with passing generic coordinates into `MomentumObject` classes. The same is under discussion for the shortcut constructors (`obj` here), which can be handled in another PR.

henryiii:
Feel free to turn of the unreachable check for mypy if it triggers too much, and 1-2 type ignores need `[misc]` added.

Saransh-cpp:
Done! I've also refactored out the type safety code from the `obj` constructor.

henryiii:
I can't approve "my own" (according to GitHub) PR, so verbally approving here!

Saransh-cpp:
Merged! Thanks for the review! 🎉

PR #95, opened 2021-04-16
Allow lowercase `e` and `m` where `E`/`energy` and `M`/`mass` are expected.

jpivarski:


jpivarski:
There's one more function that needs to be updated, but it moved in PR #94. That one will have to be merged before I can finish this.

PR #100, opened 2021-04-28
Fix '_wrap_result' for methods called on ak.Record.

jpivarski:


jpivarski:
@henryiii, for the "review," I'm just getting into a habit of being formal, since we're sharing this project. If it looks good to you, go ahead and merge it (I'm done).

This PR introduces the practice of naming tests after the issue that motivated them, but inside of a single `test_issues.py` file, rather than being the only way tests get named, as in Awkward and Uproot.

henryiii (approved):
LGTM!

PR #126, opened 2021-07-27
Added contravariant transformation method

AdvaitDhingra:
This adds a method for contravariant coordinate transformations.

This is my first real Python project, so I'm sure theres something wrong with my code as it is.

Please guide me in fixing it rather than closing the PR :D Thank you


jpivarski:
Thanks for taking the time to contribute!

As you can see from the failing tests, there are a few problems with this PR. The first is that Vector does not have SymPy as a dependency and therefore can't import it. That's not likely to change, as it would make Vector a little harder to install for anyone who might have a problem installing SymPy. Although I can't think of why that would be, we generally try to shrink-wrap a library to the smallest possible set of dependencies, just in case.

Fortunately, Python lets us cheat; if there's only a few functions that use a particular library, we can put the `import` statement for that library inside the function calls, rather than at the top of the source code file. People who doesn't use those functions don't have to install the library; those who do will get an `ImportError` that can include instructions on how to install it ([like this example](https://github.com/scikit-hep/uproot4/blob/53d739516fa702c9635495710479c0d4d88fe736/src/uproot/extras.py#L44-L57)). Equivalently, the `import` can be in a source code file that isn't imported by `src/vector/__init__.py`. The `import` line won't be executed unless something prompts the submodule to be imported.

Including SymPy, and therefore the ability to take symbolic derivatives, would not just be a new method—it would be a new backend. Currently, Vector was designed to have 3 backends: Python objects, NumPy arrays, and Awkward Arrays. But the idea has always been to leave the door open to more. We had been thinking that JAX would provide autodifferentiation, though SymPy's symbolic derivatives are distinct, with their own benefits and drawbacks. To enable this, all of the [_compute](https://github.com/scikit-hep/vector/tree/main/src/vector/_compute) functions were implemented with strict rules that forbid control structures like `if` and `for` loops (checked by [this test](https://github.com/scikit-hep/vector/blob/main/tests/test_compute_features.py)). Also, the functions (`sqrt`, `cos`, `exp`, etc.) are not taken from Python's `math` library or from `numpy`, but from an abstract `lib` that gets passed in as an argument. These are exactly the precautions that would be needed to make a SymPy backend possible—so you're in luck, if that's what you're trying to do.

However, actually adding a SymPy backend would be a significant amount of work, and I would not recommend it as a first Python project. It would require adding a new suite of classes, highlighted in gray in this figure (click to enlarge):



I don't mean for that to be foreboding, and I would happily guide an experienced Python developer through the steps involved. It looks like a lot of boxes to fill in, but they're straightforward, only needed to connect the Vector infrastructure with the relevant types from SymPy. But as a _first_ Python project? It's probably too much to take in at the beginning.

If your goal is to learn Python by contributing, there are other projects within Scikit-HEP that would provide a more gentle learning curve. Maybe introduce yourself and ask about it at https://gitter.im/Scikit-HEP/community ? The histogram projects are in rapid development, and usually a small piece can be isolated from one of those.

(I just looked at your profile: you're 16 and you know about covariant/contravariant derivatives? I'm impressed! And now I've clicked around your repos. You have quite a lot of mathematical sophistication, which helps a lot when contributing to scientific software. Please do get in touch on https://gitter.im/Scikit-HEP/community and someone in our community—maybe me, if I think about it a bit more—can come up with some projects that are mutually beneficial, solving some problems for us and giving you experience and projects you can refer to when applying for college.)

AdvaitDhingra:
@jpivarski thank you so much for your kind words. I understand that this isn't the best way to learn Python. Maybe I can try to find a way to get the result with numpy to simply things.

I'll be sure to swing by the forum 😀

jpivarski:
I'm closing this now because we've talked about it and adding a SymPy backend would be a major project.

I'll be on the lookout for other ways you can contribute, and I hope the other Scikit-HEP developers have ideas as well.

Issue #127, opened 2021-07-29
Allow overriding of protocol methods in subclasses


Sinclert:
Hi there 👋🏻 

I opened this issue to discuss one particular use case we, at [MadMiner](https://github.com/diana-hep/madminer), are having trouble with.

We have a custom vector class, [MadMinerParticle](https://github.com/diana-hep/madminer/blob/2e39992dfdf15f73574a8885bd9c3bfa3a63417a/madminer/utils/particle.py#L7), overriding some of the protocol methods coming from its parent class `MomentumVector4D` (which, in reality, inherits them from `VectorObject`). When using one of these methods (our trouble is with `__iadd__`), we noticed that the operation makes use of the [\_\_array_ufunc\_\_](https://github.com/scikit-hep/vector/blob/v0.8.3/src/vector/_backends/object_.py#L238) method, which calls the [_handler_of](https://github.com/scikit-hep/vector/blob/v0.8.3/src/vector/_methods.py#L2536-L2556) dispatch function, obtaining the module in which the provided object class was defined.

https://github.com/scikit-hep/vector/blob/d9cd956072a9f9794b14a1ca5985271204d1a6dd/src/vector/_methods.py#L2547-L2553

The problem is that, with a custom `VectorObject` subclass, the module information points to a Python module that is on the client code, and therefore, not recognized among [the valid list of modules](https://github.com/scikit-hep/vector/blob/v0.8.3/src/vector/_methods.py#L2529-L2533).

https://github.com/scikit-hep/vector/blob/d9cd956072a9f9794b14a1ca5985271204d1a6dd/src/vector/_methods.py#L2529-L2533

---

Could you expand the logic on the [_handler_of](https://github.com/scikit-hep/vector/blob/v0.8.3/src/vector/_methods.py#L2536-L2556) dispatch function so that, not only the object's type `__module__` gets looked up, but the complete list of parent classes in MRO order?

Thanks in advance.

jpivarski:
It's not clear to me what you're asking, but if you know how you want it to change, could you try that as a pull request? (It doesn't help that it's been a while since I've looked at this code, but you seem to know exactly what the problem is.)

When I wrote it, I wasn't thinking about what would happen if these classes are subclassed by other libraries. If this turns out to be going against the grain, what about using [composition, rather than inheritance](https://en.wikipedia.org/wiki/Composition_over_inheritance)? (Just as an idea to consider.)

Sinclert:
[...]

PR #128, opened 2021-07-29
feat: allow overriding of protocol methods in subclasses

Sinclert:
This PR addresses issue https://github.com/scikit-hep/vector/issues/127 by extending the [_handler_of](https://github.com/scikit-hep/vector/blob/5ba111555bae20887e417c1e73101a940d06fc57/src/vector/_methods.py#L2536-L2556) function to consider the module of all the inherited classes of a certain object. This is neccesary to allow the definition of client-side custom classes that can inherit from the library ones, and override their protocol methods (`__iadd__`, `__isub__` ...).

### Additional clarifications
- I inverted the logic of the `if isinstance(obj, Vector)` to reduce indentation.
- I imported Python's `contextlib.suppress` helper to reduce indentation.

cc @jpivarski

### DIFF
--- a/src/vector/_methods.py
+++ b/src/vector/_methods.py
@@ -2535,16 +2535,14 @@ _handler_priority = [
 
 
 def _get_handler_index(obj: VectorProtocol) -> int:
-    """
-    Returns the index of the first valid handler checking the list of parent classes
-    Returns -1 if no handler is found
-    """
+    """Returns the index of the first valid handler checking the list of parent classes"""
     index = -1
     for cls in type(obj).__mro__:
         with suppress(ValueError):
             index = _handler_priority.index(cls.__module__)
             break
 
+    assert index != -1
     return index
 
 
--- a/src/vector/_methods.py
+++ b/src/vector/_methods.py
@@ -2559,11 +2559,12 @@ def _handler_of(*objects: VectorProtocol) -> VectorProtocol:
     """
     handler = None
     for obj in objects:
-        if isinstance(obj, Vector):
-            if handler is None:
-                handler = obj
-            elif _get_handler_index(obj) > _get_handler_index(handler):
-                handler = obj
+        if not isinstance(obj, Vector):
+            continue
+        if handler is None:
+            handler = obj
+        elif _get_handler_index(obj) > _get_handler_index(handler):
+            handler = obj
 
     assert handler is not None
     return handler
--- a/src/vector/_methods.py
+++ b/src/vector/_methods.py
@@ -2539,7 +2539,9 @@ def _get_handler_index(obj: VectorProtocol) -> int:
     for cls in type(obj).__mro__:
         with suppress(ValueError):
             return _handler_priority.index(cls.__module__)
-    raise AssertionError(f"Could not find a valid handler for {obj}! This should not happen.")
+    raise AssertionError(
+        f"Could not find a valid handler for {obj}! This should not happen."
+    )
 
 
 def _handler_of(*objects: VectorProtocol) -> VectorProtocol:
--- a/src/vector/_methods.py
+++ b/src/vector/_methods.py
@@ -5,6 +5,7 @@
 
 import typing
 
+from contextlib import suppress
 from vector._typeutils import (
     BoolCollection,
     ScalarCollection,
@@ -2533,6 +2534,20 @@ _handler_priority = [
 ]
 
 
+def _get_handler_index(obj: VectorProtocol) -> int:
+    """
+    Returns the index of the first valid handler checking the list of parent classes
+    Returns -1 if no handler is found
+    """
+    index = -1
+    for cls in type(obj).__mro__:
+        with suppress(ValueError):
+            index = _handler_priority.index(cls.__module__)
+            break
+
+    return index
+
+
 def _handler_of(*objects: VectorProtocol) -> VectorProtocol:
     """
     Determines which vector should wrap the output of a dispatched function.
@@ -2547,9 +2562,7 @@ def _handler_of(*objects: VectorProtocol) -> VectorProtocol:
         if isinstance(obj, Vector):
             if handler is None:
                 handler = obj
-            elif _handler_priority.index(
-                type(obj).__module__
-            ) > _handler_priority.index(type(handler).__module__):
+            elif _get_handler_index(obj) > _get_handler_index(handler):
                 handler = obj
 
     assert handler is not None
--- a/src/vector/_methods.py
+++ b/src/vector/_methods.py
@@ -2536,14 +2536,10 @@ _handler_priority = [
 
 def _get_handler_index(obj: VectorProtocol) -> int:
     """Returns the index of the first valid handler checking the list of parent classes"""
-    index = -1
     for cls in type(obj).__mro__:
         with suppress(ValueError):
-            index = _handler_priority.index(cls.__module__)
-            break
-
-    assert index != -1
-    return index
+            return _handler_priority.index(cls.__module__)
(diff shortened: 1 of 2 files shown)


jpivarski:
I personally think that indented control structures are more clear than imperative-jumping logic like `break` and `continue`, but I'm fine with it being written like this. I checked on [contextlib.suppress](https://docs.python.org/3/library/contextlib.html#contextlib.suppress) and it was introduced in Python 3.4, but Vector is a Python 3-only library, so that's fine.

~~Now, thinking about the substance of this: what this change does is it makes instances of any subclasses of VectorObject, VectorNumpy, or VectorAwkward be "lower priority" than any of these known types. That means that when a CustomVector is used in an expression with any other kind of vector, that other vector is used to compute it, regardless of whether CustomVector is derived from VectorObject, VectorNumpy, or VectorAwkward. That's probably not what you want. The reason we have these priority rules is so that `single_vector_object + array_of_vectors` gets computed by the `array_of_vectors`'s methods, which broadcasts the `single_vector_object` to every item in the array. If it's an array-of-vectors class that gets inherited, you'd want that derived class to win over VectorObject, so that the VectorObject gets broadcasted.~~

~~What you originally described in issue #127 would be better: walk the `.mro` of the class object, checking to see if its `__module__` is in `_handler_priority`, and stop when you find it, returning that as its priority. That way, a CustomSingleVector inheriting from VectorObject would have lower priority than CustomArrayOfVectors inheriting from VectorNumpy or VectorAwkward, and you can be sure to get the right broadcasting behavior from the array classes.~~

Nevermind: that's what you're doing already. I was misled by the "return -1 if not handler is found"—that can't happen because `_get_handler_index` is called on instances of `VectorProtocol`; they _must_ have a VectorObject, VectorNumpy, or VectorAwkward (or other, future backends) in their inheritance tree by construction, so you could `assert index != -1` before returning. Objects that aren't `Vector`s are `continue`d before `_get_handler_index` can be called on them.

So if you think this PR is ready to merge, I am. (I'd only change the docstring of `_get_handler_index` to not mention the `-1` case and maybe add an assertion for it.)

jpivarski:
Thanks!

Sinclert:
You are completely right: if the `_get_handler_index` function is only called within the `_handler_of` function for-loop, then all objects are going to be a `Vector` or some sort.

I addressed your comment on https://github.com/scikit-hep/vector/pull/128/commits/9c60f5fa5e724a7b59b2fdaf8d5fe24acc18b1ed, and I think it is ready to merge.

src/vector/_methods.py
```diff
@@ -2533,6 +2534,18 @@ def _from_signature(
...
+def _get_handler_index(obj: VectorProtocol) -> int:
+    """Returns the index of the first valid handler checking the list of parent classes"""
+    index = -1
+    for cls in type(obj).__mro__:
+        with suppress(ValueError):
+            index = _handler_priority.index(cls.__module__)
+            break
+
+    assert index != -1
+    return index
```
henryiii, line 2546:
```suggestion
    for cls in type(obj).__mro__:
        with suppress(ValueError):
            return _handler_priority.index(cls.__module__)
    raise IndexError(f"Could not find a valid handler for {obj}! This should not happen.")

    assert index != -1
    return index
```

jpivarski: Shouldn't it be an AssertionError, since it would be an internal logic error, not a user error?

Otherwise, this implementation is cleaner (no need to introduce the `-1`).

src/vector/_methods.py
```diff
@@ -2544,13 +2557,12 @@ def _handler_of(*objects: VectorProtocol) -> VectorProtocol:
...
     for obj in objects:
-        if isinstance(obj, Vector):
-            if handler is None:
-                handler = obj
-            elif _handler_priority.index(
-                type(obj).__module__
-            ) > _handler_priority.index(type(handler).__module__):
-                handler = obj
+        if not isinstance(obj, Vector):
+            continue
```
henryiii, line 2561: This is "guard" style, and I have come to like it. While the continue is slightly irritating, it simplifies the logic for the reader. This is "guarded", that is, you read "do nothing if this is not a vector", while the nested block before requires you look to see there's no else block, and requires more layers (mentally, as well as visually). If you have even more deeply nested structures, this is a must.

I am less fond of the guard inside a loop as I would be elsewhere, but I do think it's an improvement.

src/vector/_methods.py
henryiii: Probably, yes. It should not be an assert, as the main reason to use assert is to have the ability to turn off the assertions, but this one should never be reached, and we do always want to throw if we reach it.

src/vector/_methods.py
henryiii: (The "normal" guard clause would be `if cond: return` at the top of a function)

src/vector/_methods.py
henryiii: (Changed in the suggestion)

src/vector/_methods.py
jpivarski: I frequently use guard style when the error case raises an exception or returns early from a function.

But personally, I find any logic involving `break` and `continue` to be less clear than fenced-off blocks (functional programming influence), though I also don't have strong enough feelings to think it ought to be changed. If you (both) think it's easier to read, that's reason enough to keep it.

Sinclert:
Just checking: would these changes produce a new minor version release? or you guys have other time-frame in mind?

I am asking because the current MadMiner approach to the subclassing is broken, and I would like to know if we should create a temporal solution, or just wait for... -check notes- `vector 0.8.4` to be released.

src/vector/_methods.py
jpivarski: But it's also true that an assertion represents something that should never happen. If there's ever a situation in which an assertion would fail, wrong behavior would result from letting it slip by. Turning off assertions trades safety for speed—which can be a reasonable thing to do when you're confident—but the above argument would apply to every assertion, not just this one, and then what would assertions be for?

(Note, I'm saying that it shouldn't be possible for this to get out of the `for` loop due to constraints applied elsewhere, not that it is possible and we just shouldn't allow it. That's why I think it's an assertion.)

If you're happy with it, I'll commit this suggestion.

jpivarski:
I think we can push a new patch release as soon as this is merged.

src/vector/_methods.py
Sinclert: I am OK with it too. Feel free to commit.

henryiii: If something is never reached, turning it off doesn't speed anything up, so there's no need to turn it off. Asserts should be optional checks, not fallbacks when nothing else works. That's why the assert statement takes a condition - it's a check, not a fallback.

Unrelated to this PR, but slightly to the general discussion, Python 3.11 will finally make a `try` block virtually no-cost if there is no throw. Currently it's less expensive than an `if` if there is no throw, but non-zero.

src/vector/_methods.py
henryiii: Those are the best places to use it. :) I think it's very close here, and I'd not probably ask for one over the other in a review, but seeing both, I do have a very minor preference for the guard.

"break" specifically is very useful in Python loops due to the else clause.

src/vector/_methods.py
henryiii: Actually, this is not the best way to do this, as handling the exception is expensive.

jpivarski: There wouldn't be a speed difference here, since (in principle) the `raise AssertionError` line is never reached. I was referring to not having to compute expressions "`xyz`" in an `assert xyz` statement in general—the only reason I can imagine for turning off assertions.

PR #131, opened 2021-08-02
Added Requirements.txt file and added it to the README

AdvaitDhingra:
I created a requirements.txt file to simplify the installation of the requirements. I also added it to the README installation guide.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ Main features of Vector:
 
 ## Installation
 
-To install, use `pip install vector` or your favorite way to instal in an environment.
+To install, first use `pip install -r requirements.txt` to install the requirements. Then use `pip install vector` or your favorite way to instal in an environment.
 
 ## Overview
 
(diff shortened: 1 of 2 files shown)


henryiii:
This is not simpler, it is much harder. A user then has to download this file from somewhere, and then run it, then we have no control over it - if we want to change a minimum or add a requirement, then the user has to find and download this file again.

For NumPy, we already list it as a `install_requires` requirement, so it always will be installed, with the correct version, when you pip install. What I think you wanted was some way to also include numba and awkward, which are optional requirements. Those would be best as `extra_requires`. In fact, if you check setup.py, we already have `dev`, which installs these two (and pytest). So a user can do `pip install "vector[dev]"` and get these, with correct versions.

I highly recommend you carefully read https://scikit-hep.org/developer which has some excellent information on these sorts of things. If you feel the docs should be updated to make "dev" a bit more obvious, we could do that - but for most users, if they are interested in using vector + awkward or vector + numba, they will already have awkward or numba.

PR #151, opened 2021-11-03
Fixed numpy array code examples in documentation

cansik:
It seems that `vector.arr` creates an `awkward` array, but the example code should create numpy arrays. This is a fix for the documentation about Numpy arrays of vectors.


jpivarski:
Thanks! I think this is something that changed.

PR #163, opened 2022-01-06
VectorNumpy pickle suppport to enable multiprocessing

cansik:
When sharing `VectorNumpy` based arrays (like `VectorNumpy3D`) between processes in python, [pickle](https://docs.python.org/3/library/pickle.html) is used for serialisation and deserialisation of the objects. But it seems that structured arrays which inherhit form a Numpy array lose part of their state during that process.

Here is an example that shows the problem:

```python
import pickle
import numpy as np
import vector

test = np.arange(0, 24, 0.1).view([("x", float), ("y", float), ("z", float)]).view(vector.VectorNumpy3D)

test_pickled = pickle.dumps(test)
test_new = pickle.loads(test_pickled)

print(test_new) # Leads to exception A
print(test_new[0]) # Leads to exception B
```

#### Exception A
First, there is a problem that `__array_finalize__` is called twice from `pickle.loads`, once with the actual data and once with a None type. This leads to the following error message:

```python
VectorNumpy3D must have a structured dtype containing fields ("x", "y") or ("rho", "phi")
```

It is possible to prevent this behaviour by checking for `None` arguments and just ignore them ([L900-L901](https://github.com/cansik/vector/blob/numpy-pickle-support/src/vector/_backends/numpy_.py#L900-L901))

#### Exception B
After that it is possible to `print` the unpickled array, but accessing the elements itself (`test_new[0]`) leads to another error message. I guess it has to do with the construction of the VectorObjects itself.

```python
TypeError: __init__() takes 3 positional arguments but 4 were given
```

#### Solution
After a bit of research I found this [stackoverflow answer](https://stackoverflow.com/a/26599346/1138326), why subclasses of numpy can lose their state when being pickled. Implementing this in the `VectorNumpy` parent class ([L433-L440](https://github.com/cansik/vector/blob/numpy-pickle-support/src/vector/_backends/numpy_.py#L433-L440)) fixes it also for all subclasses like `VectorNumpy3D`. To check if the implementation works, there are six new tests pickling/unpickling every Vector/Momentum-Numpy array variant.

In my own projects the code is already used for multi-processed calculations and result-sharing in the `VectorNumpy4D` format. Would be great if you could review it and give me feedback if this implementation is ok.

### DIFF
--- a/src/vector/_backends/numpy_.py
+++ b/src/vector/_backends/numpy_.py
@@ -431,13 +431,13 @@ class VectorNumpy(Vector, GetItem):
         return numpy.not_equal(self, other)  # type: ignore
 
     def __reduce__(self):
-        pickled_state = super(VectorNumpy, self).__reduce__()
+        pickled_state = super().__reduce__()
         new_state = pickled_state[2] + (self.__dict__,)
         return (pickled_state[0], pickled_state[1], new_state)
 
     def __setstate__(self, state):
         self.__dict__.update(state[-1])
-        super(VectorNumpy, self).__setstate__(state[0:-1])
+        super().__setstate__(state[0:-1])
 
     def __array_ufunc__(
         self,
--- a/src/vector/_backends/numpy_.py
+++ b/src/vector/_backends/numpy_.py
@@ -430,6 +430,15 @@ class VectorNumpy(Vector, GetItem):
     def __ne__(self, other: typing.Any) -> typing.Any:
         return numpy.not_equal(self, other)  # type: ignore
 
+    def __reduce__(self):
+        pickled_state = super(VectorNumpy, self).__reduce__()
+        new_state = pickled_state[2] + (self.__dict__,)
+        return (pickled_state[0], pickled_state[1], new_state)
+
+    def __setstate__(self, state):
+        self.__dict__.update(state[-1])
+        super(VectorNumpy, self).__setstate__(state[0:-1])
+
     def __array_ufunc__(
         self,
         ufunc: typing.Any,
@@ -665,6 +674,9 @@ class VectorNumpy2D(VectorNumpy, Planar, Vector2D, numpy.ndarray):  # type: igno
         return array.view(cls)
 
     def __array_finalize__(self, obj: typing.Any) -> None:
+        if obj is None:
+            return
+
         if _has(self, ("x", "y")):
             self._azimuthal_type = AzimuthalNumpyXY
         elif _has(self, ("rho", "phi")):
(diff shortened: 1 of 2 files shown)


henryiii:
I'd usually use setstate/getstate (reduce should only be used if getstate cannot be used). Was there a reason getstate could not be used or would be significantly more complex? Given you are just slightly modifying a super call to reduce, I think reduce is pretty likely to be safe in this case.

Sort of related, but nothing to do directly with this PR: can we add `__slots__`?

henryiii (approved):
I'm happy with this, ideally you can commend on `__reduce__`, but as long as it's clearly the best solution, I'm happy.

src/vector/_backends/numpy_.py
```diff
@@ -430,6 +430,15 @@ def __eq__(self, other: typing.Any) -> typing.Any:
...
         return numpy.not_equal(self, other)  # type: ignore
 
+    def __reduce__(self) -> typing.Union[str, typing.Tuple[typing.Any, ...]]:
+        pickled_state = super().__reduce__()
+        new_state = (*pickled_state[2], self.__dict__)
+        return pickled_state[0], pickled_state[1], new_state
+
+    def __setstate__(self, state: typing.Any) -> None:
+        self.__dict__.update(state[-1])
+        super().__setstate__(state[0:-1])  # type: ignore
```
henryiii, line 440: Reminder to myself: we should enable `--show-error-codes` on mypy and require the error codes here, so we know why these are being ignored. I can do that after the PR.

jpivarski:
> Sort of related, but nothing to do directly with this PR: can we add `__slots__`?

We could use `__slots__`. For the most part, I haven't had time to focus on Vector apart from keeping up with the most urgent bug reports (i.e. calling a function returns the wrong answer).

From what I know, `__slots__` would be an optimization, but it might have other benefits you know of. It could be applied to `Vector/MomentumNumpy*D` and `Vector/MomentumObject*D`, but not the Awkward Array backend because that one doesn't have extra instance state.

jpivarski:
@cansik Give us an indication of whether you're done and Henry or I will squash-and-merge your PR. Thanks!

henryiii:
My favorite reason to add `__slots__` is that it causes `v.misspelled = 2` to start failing. If you still want a dict, it can give you extra control over the `__dict__` by explicitly adding it to `__slots__`; then you can have things in `__slots__` and things in `__dict__`.

cansik:
@jpivarski & @henryiii Thanks for the review. I mainly followed the explanation in the [stackoverflow answer](https://stackoverflow.com/questions/26598109/preserve-custom-attributes-when-pickling-subclass-of-numpy-array/26599346#26599346) and it seems that numpy does not support getstate, but reduce by default: https://stackoverflow.com/a/54487727/1138326

For this reason, I would leave the implementation as it is and I have nothing to add to it.

jpivarski:
> numpy does not support getstate

That's definitely a good enough reason. NumPy's restrictions on subclasses are pretty tight.

Thanks, I'll merge it now!

PR #172, opened 2022-02-18
fix bad values for high (abs) eta

bfis:
The current eta calculation gives very wrong values for higher (absolute) eta values. This effect is more prominent if the underlying values are `float32`. The issue lies in the used formula, which adds rho and z, which may be more orders of magnitude apart than the float numbers can represent. The solution is to use a different way (mathematically equivalent) which has greater numerical stability. The same method is [used in ROOT](https://github.com/root-project/root/blob/30ab704cbe645d2b2f36cb071905e462771df386/math/genvector/inc/Math/GenVector/eta.h#L59).

The behavior introduced in #139 is retained (for the most part). Otherwise, the implementation is closer to ROOT how it handles infinities. It differs from ROOT as follow:
- for rho=0 (or NaN) it does not introduce (somewhat arbitrary) "maximum" values and yields infinities (or NaN) instead
- it doesn't perform an `sqrt` approximation (via Taylor series) for small (abs) eta
- it retains the sign (of z) for very small eta

Testing should be covered by #36. 

<details>
<summary>Detailed difference between old, new, and ROOT behaviour.</summary>

The following tables were produces by this code:
```python
import numpy as np
from vector._compute.spatial.eta import rhophi_z
import ROOT

@np.vectorize
def root_eta(rho, z):
    return ROOT.Math.PxPyPzEVector(np.abs(rho), 0, z, np.nan).Eta()

v = np.array([-np.inf,  -1e20, -1.,  -0.,  np.nan,   0.,   1.,  1e20, np.inf])

# eval via:
rhophi_z(v[:, None], 0, v[None, :])
root_eta(v[:, None], v[None, :])
```

Axes: 
horizontal: z
vertical: rho

### old
```
array([[ 0.000e+000, -0.000e+000, -0.000e+000, -0.000e+000,         nan,  0.000e+000,  0.000e+000,  0.000e+000,  0.000e+000],
       [ 0.000e+000, -8.814e-001, -1.000e-020, -0.000e+000,         nan,  0.000e+000,  1.000e-020,  8.814e-001,  0.000e+000],
       [ 0.000e+000, -1.798e+308, -8.814e-001, -0.000e+000,         nan,  0.000e+000,  8.814e-001,  1.798e+308,  0.000e+000],
       [ 0.000e+000, -1.798e+308, -1.798e+308,  0.000e+000,         nan,  0.000e+000,  1.798e+308,  1.798e+308,  0.000e+000],
       [ 0.000e+000,  0.000e+000,  0.000e+000,  0.000e+000,         nan,  0.000e+000,  0.000e+000,  0.000e+000,  0.000e+000],
       [ 0.000e+000, -1.798e+308, -1.798e+308,  0.000e+000,         nan,  0.000e+000,  1.798e+308,  1.798e+308,  0.000e+000],
       [ 0.000e+000, -1.798e+308, -8.814e-001, -0.000e+000,         nan,  0.000e+000,  8.814e-001,  1.798e+308,  0.000e+000],
       [ 0.000e+000, -8.814e-001, -1.000e-020, -0.000e+000,         nan,  0.000e+000,  1.000e-020,  8.814e-001,  0.000e+000],
       [ 0.000e+000, -0.000e+000, -0.000e+000, -0.000e+000,         nan,  0.000e+000,  0.000e+000,  0.000e+000,  0.000e+000]])
```

### new
```
array([[        nan, -0.000e+100, -0.000e+000, -0.000e+000,         nan,  0.000e+000,  0.000e+000,  0.000e+000,         nan],
       [       -inf, -8.814e-001, -1.000e-020, -0.000e+000,         nan,  0.000e+000,  1.000e-020,  8.814e-001,         inf],
       [       -inf, -4.674e+001, -8.814e-001, -0.000e+000,         nan,  0.000e+000,  8.814e-001,  4.674e+001,         inf],
       [       -inf,        -inf,        -inf, -0.000e+000,         nan,  0.000e+000,         inf,         inf,         inf],
       [        nan,         nan,         nan, -0.000e+000,         nan,  0.000e+000,         nan,         nan,         nan],
       [       -inf,        -inf,        -inf, -0.000e+000,         nan,  0.000e+000,         inf,         inf,         inf],
       [       -inf, -4.674e+001, -8.814e-001, -0.000e+000,         nan,  0.000e+000,  8.814e-001,  4.674e+001,         inf],
       [       -inf, -8.814e-001, -1.000e-020, -0.000e+000,         nan,  0.000e+000,  1.000e-020,  8.814e-001,         inf],
       [        nan, -0.000e+000, -0.000e+000, -0.000e+000,         nan,  0.000e+000,  0.000e+000,  0.000e+000,         nan]])


```

### ROOT
```
array([[       nan,  0.000e+00,  0.000e+00,  0.000e+00,        nan,  0.000e+00,  0.000e+00,  0.000e+00,        nan],
       [      -inf, -8.814e-01,  0.000e+00,  0.000e+00,        nan,  0.000e+00,  0.000e+00,  8.814e-01,        inf],
       [      -inf, -4.674e+01, -8.814e-01,  0.000e+00,        nan,  0.000e+00,  8.814e-01,  4.674e+01,        inf],
       [      -inf, -1.000e+20, -2.276e+04,  0.000e+00,        nan,  0.000e+00,  2.276e+04,  1.000e+20,        inf],
       [      -inf, -1.000e+20, -2.276e+04,  0.000e+00,        nan,  0.000e+00,  2.276e+04,  1.000e+20,        inf],
       [      -inf, -1.000e+20, -2.276e+04,  0.000e+00,        nan,  0.000e+00,  2.276e+04,  1.000e+20,        inf],
       [      -inf, -4.674e+01, -8.814e-01,  0.000e+00,        nan,  0.000e+00,  8.814e-01,  4.674e+01,        inf],
       [      -inf, -8.814e-01,  0.000e+00,  0.000e+00,        nan,  0.000e+00,  0.000e+00,  8.814e-01,        inf],
       [       nan,  0.000e+00,  0.000e+00,  0.000e+00,        nan,  0.000e+00,  0.000e+00,  0.000e+00,        nan]])
```

</details>


jpivarski:
By the way, I'm looking at this. This would be the first `compute` function to use `lib.where`, and that might be a problem in Numba. Reading the formula, it looks like it's not necessary: `z` is in the numerator of an `arcsinh`, and all you're doing when you replace the `z=0` cases with `z` is you're replacing `0`. Oh, unless the `x` and `y` cause a NaN. Let me think about that...

jpivarski:
Yeah, there were two errors, and you've fixed the non-Numba one. Removing `lib.where` fixes Numba. If the `lib.where` is truly necessary, then there might be a work-around for Numba, though I'm not sure.

jpivarski:
Oh, you beat me to it. The issue with Numba is that it's a different type, which is also an issue in the plain-object backend. Numba computes the `np.where` correctly, but there's an issue downstream with having a zero-dimensional array instead of a scalar.

jpivarski:
This is a little disgusting, but what about

```python
>>> for z in (-np.inf, -1.1, 0.0, 1.1, np.inf, np.nan):
...     print(z, np.nan_to_num((z != 0) * np.inf, posinf=np.nan))
... 
-inf nan
-1.1 nan
0.0 0.0
1.1 nan
inf nan
nan nan
```

bfis:
Entirely without `where` it yields this:
```
array([[    nan,  -0.   ,  -0.   ,  -0.   ,     nan,   0.   ,   0.   ,   0.   ,     nan],
       [   -inf,  -0.881,  -0.   ,  -0.   ,     nan,   0.   ,   0.   ,   0.881,     inf],
       [   -inf, -46.745,  -0.881,  -0.   ,     nan,   0.   ,   0.881,  46.745,     inf],
       [   -inf,    -inf,    -inf,     nan,     nan,     nan,     inf,     inf,     inf],
       [    nan,     nan,     nan,     nan,     nan,     nan,     nan,     nan,     nan],
       [   -inf,    -inf,    -inf,     nan,     nan,     nan,     inf,     inf,     inf],
       [   -inf, -46.745,  -0.881,  -0.   ,     nan,   0.   ,   0.881,  46.745,     inf],
       [   -inf,  -0.881,  -0.   ,  -0.   ,     nan,   0.   ,   0.   ,   0.881,     inf],
       [    nan,  -0.   ,  -0.   ,  -0.   ,     nan,   0.   ,   0.   ,   0.   ,     nan]])
```

The NaNs for rho=NaN are acceptable, but for rho=0=z it's an issue.

jpivarski:
The "disassemble check" is intended to constrain the kinds of functions that are used in compute functions, and if it had run long enough, it would have complained about `lib.where`. In the most recent run, it complained about `lib.arcsinh`, but that should definitely be an allowed function. I've fixed that (by including all the hyperbolics).

bfis:
With your proposal it looks like this:
```
array([[    nan,  -0.   ,  -0.   ,  -0.   ,     nan,   0.   ,   0.   ,   0.   ,     nan],
       [   -inf,  -0.881,  -0.   ,  -0.   ,     nan,   0.   ,   0.   ,   0.881,     inf],
       [   -inf, -46.745,  -0.881,  -0.   ,     nan,   0.   ,   0.881,  46.745,     inf],
       [   -inf,    -inf,    -inf,   0.   ,     nan,   0.   ,     inf,     inf,     inf],
       [    nan,     nan,     nan,   0.   ,     nan,   0.   ,     nan,     nan,     nan],
       [   -inf,    -inf,    -inf,   0.   ,     nan,   0.   ,     inf,     inf,     inf],
       [   -inf, -46.745,  -0.881,  -0.   ,     nan,   0.   ,   0.881,  46.745,     inf],
       [   -inf,  -0.881,  -0.   ,  -0.   ,     nan,   0.   ,   0.   ,   0.881,     inf],
       [    nan,  -0.   ,  -0.   ,  -0.   ,     nan,   0.   ,   0.   ,   0.   ,     nan]])
```

I thinks this is ok - nobody will miss missing minus for z=-0.

jpivarski:
I'm looking at the Awkward error.

jpivarski:
The problem is that the argument passed to `nan` in `lib.nan_to_num` has to be scalar. Huh.

bfis:
> The problem is that the argument passed to `nan` in `lib.nan_to_num` has to be scalar. Huh.

This also seem to be the case for the [numba "implementation"](https://github.com/scikit-hep/vector/blob/0b9bf7ef622b634b749a0fa664da2c2a565affa1/src/vector/_backends/numba_object.py#L51-L77) ...

jpivarski:
I was wrong about `np.nan_to_num`: it can take an array as the `nan`. Maybe `ak.nan_to_num` doesn't know that, though.

```python
>>> array = np.array([1, 2, 3, np.nan, np.nan, 10])
>>> np.nan_to_num(array, nan=np.array([999, 999, 999, 123, 321, 999]))
array([  1.,   2.,   3., 123., 321.,  10.])
```

jpivarski:
https://github.com/numba/numba/issues/6857 and https://github.com/numba/numba/issues/5977

I'm seeing if we can do without these workaround-implementations now.

jpivarski:
I've fixed the Numba implementation of `np.nan_to_num` (in Vector, because core Numba still doesn't have it) and the Awkward implementation of `ak.nan_to_num` to allow the `nan`, `posinf`, `neginf` arguments to be arrays (as long as they're broadcastable). These two fixes solve all the failing tests.

Along the way, I learned that if you don't set `posinf` and `neginf` (i.e. leave them as `None`), they'll be replaced with the largest and smallest finite value. That's a surprise to me, and it means that we should explicitly set them to plus and minus infinity as you've done here. I'll do that in another PR.

Unfortunately, this PR is going to be blocked until I publish a non-release candidate version of Awkward Array on PyPI. That's overdue, anyway, so I should focus on that.

jpivarski:
> I'll do that in another PR.

#173.

jpivarski:
I don't know what you did, but it has the right type:

```python
>>> import vector
>>> xyz = vector.obj(x=1.1, y=2.2, z=3.3)
>>> xyz.theta
0.6405223126794245
>>> xyz.eta
1.103586841560145
```

I'm checking more deeply now. (And I can loosen the disassemble-check to allow `lib.where` if this is the right way to go.)

jpivarski:
Multiplying by 1 turns a zero-dimensional array into a scalar. I had no idea!

```python
>>> np.where(0, 1, 1.1)
array(1.1)
>>> np.where(0, 1, 1.1) * 1
1.1
>>> type(np.where(0, 1, 1.1) * 1)
<class 'numpy.float64'>
```

While that's brilliant, I'm a little uncomfortable with the solution because new backends might not be able to handle it. I've been hoping, for instance, that we'll be able to add SymPy as a backend. `lib.nan_to_num` is something that we could justifiably make a pass-through (identity function) in SymPy, but what would be the right translation of `lib.where`? For this case, it would be "take the first argument, ignore the second" because the purpose of this `lib.where` is to handle exceptional values—we know that the "normal" values are in the first argument, "exceptional" ones are in the second. If we _always_ use `lib.where` that way, we could make that the appropriate translation...

Do the `lib.where` and `lib.nan_to_num` solutions have different extreme values? Is one set or the other more appropriate? If `lib.where` gives more appropriate extreme values, then that would be the reason for selecting it, and we'll just follow this rule that we only use it for extreme-value cleanup, and that the extreme values are always in the second argument. A disassembler-check can't enforce that; it would be a rule that we'd have to enforce manually.

henryiii:
I’d recommend checking against the new NumPy API. It’s a lot more strict, and it is supposed to be implemented by all the other libs eventually. I think it handles scalars very differently.

bfis:
@jpivarski 
> While that's brilliant, I'm a little uncomfortable with the solution because new backends might not be able to handle it. I've been hoping, for instance, that we'll be able to add SymPy as a backend. `lib.nan_to_num` is something that we could justifiably make a pass-through (identity function) in SymPy, but what would be the right translation of `lib.where`? For this case, it would be "take the first argument, ignore the second" because the purpose of this `lib.where` is to handle exceptional values—we know that the "normal" values are in the first argument, "exceptional" ones are in the second. If we _always_ use `lib.where` that way, we could make that the appropriate translation...

Sympy has the [Piecewise](https://docs.sympy.org/latest/modules/functions/elementary.html#sympy.functions.elementary.piecewise.Piecewise) function that can express the `where` behavior like this: `Piecewise((z, z == 0), (np.arcsinh(rho / z), True))`. Since this is hardly an exotic concept in math I think making use of it is appropriate.
I would not establish a convention which requires `where` to be used in a particular way, since this would be fairly error prone.
I also think that just replacing `nan_to_num` with an identity is fairly error pone too. For example, the solution prior to the `where` reintroduction ([`lib.nan_to_num((z != 0) * inf, posinf=nan)`](https://github.com/scikit-hep/vector/pull/172/commits/33f241842e788abadff6ba25037211afd07e45b6#diff-ab7d25d74d1537170e1e38ef6c342fe88c40705ea225171277ac363dfe18a824R51)) would actually yield infinities instead of zeros for the case that z=rho=0, which is very problematic.

> Do the `lib.where` and `lib.nan_to_num` solutions have different extreme values? Is one set or the other more appropriate? If `lib.where` gives more appropriate extreme values, then that would be the reason for selecting it, and we'll just follow this rule that we only use it for extreme-value cleanup, and that the extreme values are always in the second argument. A disassembler-check can't enforce that; it would be a rule that we'd have to enforce manually.

The extreme values do not chaange, merely the values for for `z=-0.` and `rho=0` or `rho=nan` would receive the correct sign. However, I still think that this is the more appropriate solution than hacking together the (practically) same effect by abusing `nan_to_num` (mutliple times).

FYI, the value matrix looks like follows (same as the `New` one in the description) and most closely follows the ROOT values sans their (questionable) quirks:
```
array([[    nan,  -0.   ,  -0.   ,  -0.   ,     nan,   0.   ,   0.   ,   0.   ,     nan],
       [   -inf,  -0.881,  -0.   ,  -0.   ,     nan,   0.   ,   0.   ,   0.881,     inf],
       [   -inf, -46.745,  -0.881,  -0.   ,     nan,   0.   ,   0.881,  46.745,     inf],
       [   -inf,    -inf,    -inf,  -0.   ,     nan,   0.   ,     inf,     inf,     inf],
       [    nan,     nan,     nan,  -0.   ,     nan,   0.   ,     nan,     nan,     nan],
       [   -inf,    -inf,    -inf,  -0.   ,     nan,   0.   ,     inf,     inf,     inf],
       [   -inf, -46.745,  -0.881,  -0.   ,     nan,   0.   ,   0.881,  46.745,     inf],
       [   -inf,  -0.881,  -0.   ,  -0.   ,     nan,   0.   ,   0.   ,   0.881,     inf],
       [    nan,  -0.   ,  -0.   ,  -0.   ,     nan,   0.   ,   0.   ,   0.   ,     nan]])
```

@henryiii 
> I’d recommend checking against the new NumPy API. It’s a lot more strict, and it is supposed to be implemented by all the other libs eventually. I think it handles scalars very differently.

I'm not familiar with what you are talking about. Is there an `where`/NumPy API variant that does handle scalars vs. rank-0 arrays consistently i.e. without unexpectedly changing from one to the other?

jpivarski:
I guess I have to say that I'm _more_ uncomfortable with the solution of multiplying the result of `np.where` by `1` to turn it from a zero-dimensional array into a scalar. For a library so concerned with portability across different values of "`lib`", that's relying pretty strongly on what looks like a NumPy quirk. I don't know how to test the new API, either, but this is exactly the sort of thing CuPy likes to break, for instance. (Even if they're in the wrong for doing so...)

It's a little ugly that `np.nan_to_num` has to be called twice, but it's being used in exactly the way it's intended: to replace extreme values. So I'd much rather go back to commit 12ae4d6c7117f85c18b67bfe6ddffcd3fb391e26, which has `eta`'s 6 implementations looking like this:

https://github.com/scikit-hep/vector/blob/12ae4d6c7117f85c18b67bfe6ddffcd3fb391e26/src/vector/_compute/spatial/eta.py#L31-L62

When the tests run again, they'll pick up Awkward 1.8.0 and it should pass. Then I'll accept the PR.

Thanks!

henryiii:
Scalars were dropped from the new API. See https://numpy.org/neps/nep-0047-array-api-standard.html.

jpivarski:
It looks like the [Array API](https://data-apis.org/array-api/latest/index.html) is now a fully drafted standard, NumPy's adoption of it ([NEP 47](https://numpy.org/neps/nep-0047-array-api-standard.html)) is under consideration, and so are the promotion rules for Python scalars ([NEP 50](https://25105-908607-gh.circle-artifacts.com/0/doc/neps/_build/html/nep-0050-scalar-promotion.html)), which is partly inspired by a [NumPy users poll](https://discuss.scientific-python.org/t/poll-future-numpy-behavior-when-mixing-arrays-numpy-scalars-and-python-scalars/202) and also [defined by the Array API](https://data-apis.org/array-api/latest/API_specification/type_promotion.html#mixing-arrays-with-python-scalars). Interesting times!

So NumPy will be dropping the distinction between zero-dimensional arrays and scalars (as CuPy already is; I guess they had some say in the Array API!). This would make the `* 1` trick still work but be redundant. Noted. But sticking to fewer functions will make it easier to add new backends, so I'd still like to go with the `nan_to_num` version.

Thanks, @bfis, for reverting it.

henryiii:
The Array API is implemented in experimental form in NumPy 1.22, use `import numpy.array_api as xp` (or whatever prefix you like using).

PR #173, opened 2022-02-18
Explicitly set posinf and neginf in nan_to_num so they stay infinite.

jpivarski:
If you don't set `posinf` and `neginf` in `np.nan_to_num`, infinite (non-NaN) values get replaced by extreme finite values, and that wasn't my intention. I only wanted to set the NaN values with the `nan` argument.


Saransh-cpp:
I think this is ready for merging now, @jpivarski.

jpivarski:
Great, thanks!

PR #183, opened 2022-04-07
Add Conda and Zenodo badges to the README

eduardo-rodrigues:
I also rearranged a bit to follow the looks of the badges as in other packages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,19 +1,20 @@
 <img alt="Vector logo" width="433" src="https://raw.githubusercontent.com/scikit-hep/vector/main/docs/_images/vector-logo.png"/>
 
-# Vector
+# Vector: arrays of 2D, 3D, and Lorentz vectors
+
+[![Scikit-HEP][sk-badge]][sk-link]
+[![PyPI version][pypi-version]][pypi-link]
+[![PyPI platforms][pypi-platforms]][pypi-link]
+[![Conda latest release][conda-version]][conda-link]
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,19 +1,20 @@
 <img alt="Vector logo" width="433" src="https://raw.githubusercontent.com/scikit-hep/vector/main/docs/_images/vector-logo.png"/>
 
-# Vector
+# Vector: arrays of 2D, 3D, and Lorentz vectors
+
+[![Scikit-HEP][sk-badge]][sk-link]
```
henryiii, line 5: The Scikit-HEP badge is always last, I believe. At least it's always in the last group. Actually, if the order of these two grapes was swapped, I think it would be pretty close. The idea was to have the standard "status" badges in the first row (like tests and such), and informational badges (like versions and DOI's) in the second row.

eduardo-rodrigues: Not sure I like, but it does not matter anyway. I prefer the way it is in Particle and DecayLanguage, which is what I followed.

Approved from you, then?

henryiii: Yes. I like the "status" badges on the top, since that's what the badges originally were for, and that's what users are used to seeing them used as. We've sort of "hijacked" them over time to show static links, but they were originally live statuses for CI. The lower/second line can be the "informational" one - expecially the scikit-hep badge, which is just a completely static link. Placing it completely last does give it a special spot, though.

Eventually we should add some recommendations and maybe even checking for the badges and their order...

PR #175, opened 2022-02-21
Implements deltaRapidityPhi and deltaRapidityPhi2.

jpivarski:
@raymondEhlers, please add tests!

PR #187, opened 2022-04-22
Tests and docs for deltaRapidityPhi

raymondEhlers:
Reopening PR after rebasing to main per discussion [here](https://github.com/scikit-hep/vector/pull/175#issuecomment-1106602474)


jpivarski (approved):
Thanks!

PR #192, opened 2022-05-21
Remove underscores

### LINKED ISSUE #136: Remove underscores from module names
Just before Vector went public, we (@henryiii and I) chose to add underscores on the module names that are not expected to be used by data analysts.

Prepended underscores in `_backends` and `_compute` were for hiding internals; appended underscores in `object_`, `numpy_`, and `awkward_` were for avoiding Python built-in and module name ambiguity, but these aren't really needed because

   * `vector.object_` would never be confused with `object` if we don't `import object from vector` (why would anyone do that?)
   * this is a purely Python 3 project; my apprehension about nested submodules with the same name as top-level modules is obsolete.

Back to the prepended underscores: `_backends` has a lot of useful classes for third party developers. For instance, I foresee @nsmith- needing to access quite a bit in `vector._backends.awkward_` in order to move NanoEvents to use Vector.

I propose

   1. to remove the appended underscores because they're not needed and ugly and
   2. to remove the prepended underscore from `_backends`. I still don't see any reason why even a third-party developer would need to see inside `_compute`.

So @henryiii, what do you think?

Saransh-cpp:
Closes #136 

## Underscores removed
- `object_` -> `object`
- `awkward_` -> `awkward`
- `numpy_` -> `numpy`
- `numba_` -> `numba`
- `_backends` -> `backends`


Saransh-cpp:
Is there a file where I should add a warning for this? The issue mentions that the `_backends` package is being used outside of vector; would this be a breaking change?

jpivarski (changes requested):
All of the modules in `backends` that I could see make good use of leading underscores for private implementations, except for the `numba` and `numba_object` backends. So just make these

   * `backends/numba` → `backends/_numba`
   * `backends/numba_objects` → `backends/_numba_objects`

and that should be it. 100% of the Numba code is registering things in Numba itself: the variables and functions that are left around after doing that are of no interest to third-party developers and should be hidden so that we're free to change them.

jpivarski (approved):

Issue #194, opened 2022-05-25
`VectorNumpy3D`'s `azimuthal` and `longitudinal` properties throw an error (similar for `VectorNumpy4D`)


Saransh-cpp:
## Reproducible example
```py
import vector

vec = vector.array(
    [
        (1.1, 2.1, 3.1),
        (1.2, 2.2, 3.2),
        (1.3, 2.3, 3.3),
        (1.4, 2.4, 4.4),
        (1.5, 2.5, 5.5)
    ], dtype=[("x", float), ("y", float), ("z", float)]
)
print(vec.azimuthal)
```
Similarly for -
- `vec.longitudinal`
- `vec.azimuthal` for a 4D NumPy vector (`VectorNumpy4D`)
- `vec.longitudinal` for a 4D NumPy vector (`VectorNumpy4D`)
- `vec.temporal` for a 4D NumPy vector (`VectorNumpy4D`)

## Error
```
>>> vec.azimuthal
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "D:\extras\IRIS-HEP\.env\lib\site-packages\numpy\core\arrayprint.py", line 1488, in _array_repr_implementation
    lst = array2string(arr, max_line_width, precision, suppress_small,
  File "D:\extras\IRIS-HEP\.env\lib\site-packages\numpy\core\arrayprint.py", line 736, in array2string
    return _array2string(a, options, separator, prefix)
  File "D:\extras\IRIS-HEP\.env\lib\site-packages\numpy\core\arrayprint.py", line 513, in wrapper
    return f(self, *args, **kwargs)
  File "D:\extras\IRIS-HEP\.env\lib\site-packages\numpy\core\arrayprint.py", line 546, in _array2string
    lst = _formatArray(a, format_function, options['linewidth'],
  File "D:\extras\IRIS-HEP\.env\lib\site-packages\numpy\core\arrayprint.py", line 889, in _formatArray
    return recurser(index=(),
  File "D:\extras\IRIS-HEP\.env\lib\site-packages\numpy\core\arrayprint.py", line 853, in recurser
    word = recurser(index + (-1,), next_hanging_indent, next_width)
  File "D:\extras\IRIS-HEP\.env\lib\site-packages\numpy\core\arrayprint.py", line 799, in recurser
    return format_function(a[index])
  File "D:\extras\IRIS-HEP\.env\lib\site-packages\vector\_backends\numpy_.py", line 218, in __getitem__
[...]

PR #195, opened 2022-05-26
docs: add docstrings in the `backends.numpy` module

Saransh-cpp:
I have tried to add docstrings for both developers and users. #194 blocked the addition of examples to some properties, but I will definitely add them once the issue is closed. I have skipped some functions and classes as I couldn't properly understand what they were doing. These are -
- `def _getitem`
- `def _setitem`
- `class GetItem`

Please let me know if I should add some more context in the docstrings or if any changes are required!

**Note**: The class variables are rendered in the documentation with their values automatically. Hence, I decided not to add docstrings for them, but I would be more than happy to override the automatically rendered stuff!

### DIFF
--- a/src/vector/backends/numpy.py
+++ b/src/vector/backends/numpy.py
@@ -209,7 +209,7 @@ def _has(
         bool: If the attribute exists or not.
 
     Examples:
-        >>> from vector._backends.numpy_ import _has
+        >>> from vector.backends.numpy import _has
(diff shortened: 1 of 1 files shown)


henryiii (commented):
`GetItem` is a mixin that provides the infrastructure for `[]` access. These are implementation details so don't think it's critical to have good docs for them. The rest is an improvement!

henryiii (approved):

PR #199, opened 2022-05-31
ci: fix a test and update CI to catch errors regularly

Saransh-cpp:
* Fixed a numerical error in tests (no idea how it came into existence suddenly, a change in the upstream repositories?)
* Added a cron job to CI
* Added workflow_dispatch to CI
* CI now cancels or skips intermediate builds


Saransh-cpp:
The coverage currently is not visible on Codecov. The link gives a "Sorry, we couldn't find that page" error, but the cause of this might be the absence of a reference (main branch) coverage report?

jpivarski (approved):
Checkmark from me; adding Codecov is a good idea. I'm not sure that the tests need to run every day, but I'm not against it.

.github/workflows/ci.yml
```diff
@@ -6,6 +6,15 @@ on:
     branches:
     - main
     - develop
+  schedule:
+    - cron: '0 2 * * *'  # Daily at 2 AM UTC
```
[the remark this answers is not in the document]
Saransh-cpp: Now that I think of it, maybe not. I added a cron job as this particular failure was detected by the `pre-commit` bot, but this error will not result in vector not functioning for users. Should I revert this?

Saransh-cpp:
Should I (squash and) merge this PR? (A general query - should I merge PRs after they are approved, or will it be performed by the maintainers? Also, should I wait for both approvals, or will one suffice to make it ready for merging?)

henryiii:
I'm fine with one approval. I'd say "anyone can merge once we get one approval, except the person that approved it". Really small PRs like a tiny bug fix can be merged by the approver.

henryiii:
(This is the sort of thing I like, so I'm reviewing it too now)

.github/workflows/ci.yml
```diff
@@ -6,6 +6,13 @@ on:
     branches:
     - main
     - develop
+  workflow_dispatch:
+
+concurrency:
+  # Skip intermediate builds: always.
+  # Cancel intermediate builds: only if it is a pull request build.
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ startsWith(github.ref, 'refs/pull/') }}
```
henryiii, line 15: I generally use true here, but this is fine too. We don't really need/use historical every-PR build status.

setup.py
```diff
@@ -10,6 +10,7 @@
     "awkward": ["awkward>=1.2.0"],
     "test": [
         "pytest>=4.6",
+        "pytest-cov==3.0.0",
```
henryiii, line 13: This shouldn't be pinned this tightly. Maybe cap at <4; leaving uncapped is fine too. Also we should go ahead and make pytest>=6 while we are at it.

tests/backends/test_awkward.py
```diff
@@ -78,27 +78,25 @@ def test_projection():
...
-            {
-                "rho": 6.4031242374328485,
-                "phi": 0.8960553845713439,
-                "eta": 0.8361481196083128,
-                "tau": 0,
-                "wow": 123,
-            }
-        ],
-    ]
+    assert out.tolist()[0][0] == pytest.approx(
```
henryiii, line 81: Wondering if I knew `approx` worked on a dict. :) Anyway, this now doesn't test completeness, this would now pass if there was one more item. Maybe

`(a, b, c), = out.to_list()`, then check `a`, `b` and `c`?

henryiii: Sorry, I think I read that wrong, I think it's `(a,), b, (c,) =`.

setup.py
Saransh-cpp: Thank you for linking the blog! Amazing read! Never thought about capping so much :)

tests/backends/test_awkward.py
Saransh-cpp: Oh yes, thanks!

PR #210, opened 2022-06-23
Add type checks in constructors

Saransh-cpp:
See #205 

Added type checks in all the existing constructors.


src/vector/backends/awkward_constructors.py
```diff
@@ -256,11 +271,20 @@ def Array(*args: typing.Any, **kwargs: typing.Any) -> typing.Any:
     import vector
     import vector.backends.awkward  # noqa: 401
 
+    if isinstance(args[0], dict):
+        if all(isinstance(x, list) for x in args[0].values()):
+            raise TypeError(
+                "Use vector.zip to construct vectors using a list of coordinates"
+            )
```
jpivarski, line 278: See above: this iterates over the data in Python. Besides not being general (it's only catching the case of dicts of lists, not array inputs, and not even non-list iterables), we only want to iterate over the data once.

To make a point about checking type metadata versus iterating over the data, checking each instance, consider this:

```python
In [1]: import awkward._v2 as ak   # I'm doing this in v2 for the ak.from_parquet function

In [2]: some_data = ak.from_parquet("https://pivarski-princeton.s3.amazonaws.com
   ...: /chep-2021-jagged-jagged-jagged/zlib9-jagged1.parquet", row_groups=[0])

In [3]: %%timeit
   ...: isinstance(ak.type(some_data.layout), ak.types.ListType)
   ...: 
   ...: 
104 µs ± 430 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

In [4]: some_data[0]
Out[4]: <Array [-0.423, 2.34, -0.757, ..., -0.297, 0.597, -0.298] type='11 * float32'>

In [5]: isinstance(some_data[0], list)
Out[5]: False

In [6]: from collections.abc import Iterable

In [7]: isinstance(some_data[0], Iterable)
Out[7]: True

In [8]: %%timeit
   ...: all(isinstance(x, Iterable) for x in some_data)
   ...: 
   ...: 
9.98 s ± 48.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [9]: len(some_data)
Out[9]: 1342176
```

It's a reasonably large array, just over a million entries, but you can check the type metadata almost _a million times faster_ than iterating over the data: 104 µs versus 9.98 s. (Note: if you're given builtin types like lists and dicts, the iteration is faster, such that checking type metadata is merely a thousand times faster than iterating over lists and dicts: 161 ms. But still, _a thousand times faster_.)

Of course, you might be _given_ lists and dicts, and then what can you do? You have to eventually turn it into an array, anyway (see [ak.to_layout](https://awkward-array.readthedocs.io/en/latest/_auto/ak.to_layout.html) above), but even doing that is only 2.5 seconds for a million entries. (The process of building the array involves type-checks in every entry.)

```python
In [10]: as_lists = ak.to_list(some_data)

In [11]: %%timeit
    ...: as_array_again = ak.from_iter(as_lists)
    ...: 
    ...: 
2.51 s ± 28.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Bottom line: you'll eventually have an array with type metadata. The data in the array will not be like Python objects, with unknown type. Use the type metadata!

src/vector/backends/awkward_constructors.py
```diff
@@ -256,11 +271,20 @@ def Array(*args: typing.Any, **kwargs: typing.Any) -> typing.Any:
...
+
     akarray = awkward.Array(*args, **kwargs)
     fields = awkward.fields(akarray)
 
     is_momentum, dimension, names, arrays = _check_names(akarray, fields)
 
+    if not _is_type_safe(akarray):
+        raise TypeError("A coordinate must be of the type int or float")
```
jpivarski, line 286: Python style is for exception messages to be lower-cased (unlike this) and not end in punctuation (like this). I think there might be a PEP on that.

src/vector/backends/numpy.py
```diff
@@ -70,6 +70,15 @@ def _array_from_columns(columns: typing.Dict[str, ArrayLike]) -> ArrayLike:
...
         array([(1., 1.), (2., 2.), (3., 4.)], dtype=[('x', '<f8'), ('y', '<f8')])
     """
+
+    for coordinates in columns.values():
+        if not all(
+            isinstance(x, (int, float, numpy.float64, numpy.int64)) for x in coordinates
+        ):
+            raise TypeError(
+                "A coordinate must be of the type int, float, numpy.int64, or numpy.float64"
+            )
```
jpivarski, line 80: Same thing as above, but for NumPy arrays instead of Awkward Arrays this time.

For NumPy arrays, you'll want to accept all the integer and floating point types:

```python
>>> dt = numpy.array([1, 2, 3]).dtype
>>> dt
dtype('int64')
>>> issubclass(dt.type, (numpy.integer, numpy.floating))
True
```

We want to exclude booleans, complex numbers, and datetime/timedeltas, as well as more exotic types like [structured arrays](https://numpy.org/doc/stable/user/basics.rec.html).

```python
>>> for t in [
...     numpy.bool_,
...     numpy.int8,
...     numpy.int16,
...     numpy.int32,
...     numpy.int64,
...     numpy.uint8,
...     numpy.uint16,
...     numpy.uint32,
...     numpy.uint64,
...     numpy.float16,   # not all platforms have this one
...     numpy.float32,
...     numpy.float64,
...     numpy.float128,   # not all platforms have this one
...     numpy.complex64,
...     numpy.complex128,
...     numpy.datetime64,
...     numpy.timedelta64,
... ]:
...     dt = numpy.dtype(t)
...     print(f"{dt} {issubclass(dt.type, (numpy.integer, numpy.floating))}")
... 
bool False
int8 True
int16 True
int32 True
int64 True
uint8 True
uint16 True
uint32 True
uint64 True
float16 True
float32 True
float64 True
float128 True
complex64 False
complex128 False
datetime64 False
timedelta64 True
```

jpivarski: Also, Python's `int` and `float` are not possible types for data from a NumPy array. If you're trying to capture all inputs, start by casting the input as a NumPy array and do this test on the dtype (so that the array is converted only once; this is the same advice as for Awkward Arrays).

Incidentally, if you want to check to see if something in Python is an integer or floating-point number, use `numbers.Real` (or `numbers.Integral` for integers-only).

```python
>>> isinstance(5, numbers.Real)
True
>>> isinstance(3.14, numbers.Real)
True
```

Not all "numbers" are subclasses of Python's concrete built-in types. The `numbers` module has abstract types for anything number-like. For instance,

```python
>>> isinstance(numpy.float32(3.14), float)
False
>>> isinstance(numpy.float32(3.14), numbers.Real)
True
```

src/vector/backends/numpy.py
```diff
@@ -297,6 +306,24 @@ def _shape_of(result: typing.Tuple[FloatArray, ...]) -> typing.Tuple[int, ...]:
...
+        "VectorNumpy4D",
+        "MomentumNumpy2D",
+        "MomentumNumpy3D",
+        "MomentumNumpy4D",
+        "CoordinatesNumpy",
+    ],
+) -> bool:
+
+    dtypes = [x[0] for x in array.dtype.fields.values()]  # type: ignore[union-attr]
+    allowed_dtypes = [int, float, numpy.int64, numpy.float64]
```
jpivarski, line 322: Again, you want all of NumPy's integer and floating-point types, which does not include `int` and `float`.

The `float32` type is fairly popular in HEP. We don't want that one failing.

jpivarski (changes requested):
See below.

Saransh-cpp:
Thank you for such detailed explanations, @jpivarski! It took me some time to go through all the details, but they were extremely helpful! The code snippets were also very helpful!

Saransh-cpp (commented):
The `awkward` constructors now look much better! The `numpy` type check still looks a bit messy as I encountered an edge case in the existing unit tests. Additionally, the type checks for `numpy` vectors is currently happening in the `__array_finalise__` method but would it be better to perform the check in the `__new__` method?

Saransh-cpp:
Oh, the light package does not include `awkward` as its dependency. I think I should use `numpy` dtypes for `numpy` vectors.

src/vector/backends/numpy.py
```diff
@@ -297,6 +298,14 @@ def _shape_of(result: typing.Tuple[FloatArray, ...]) -> typing.Tuple[int, ...]:
     return tuple(shape)
 
 
+def _is_type_safe(
+    array: typing.Any,
+) -> bool:
+    return issubclass(
+        array.dtype.type, (numpy.integer, numpy.floating)
+    ) and not issubclass(array.dtype.type, numpy.timedelta64)
```
Saransh-cpp, line 306: `bool` values are automatically converted to 1s and 0s be `numpy` and `numpy` vectors.

src/vector/backends/awkward_constructors.py
```diff
@@ -256,7 +287,14 @@ def Array(*args: typing.Any, **kwargs: typing.Any) -> typing.Any:
...
 
-    akarray = awkward.Array(*args, **kwargs)
+    layout = awkward.to_layout(args[0])
+    array_type = awkward.type(layout)
+    args = args[1:]
+
+    if not _is_type_safe(array_type):
+        raise TypeError("a coordinate must be of the type int or float")
+
+    akarray = awkward.Array(layout, *args, **kwargs)
```
jpivarski, line 297:
```suggestion
    akarray = awkward.Array(*args, **kwargs)
    array_type = akarray.type

    if not _is_type_safe(array_type):
        raise TypeError("a coordinate must be of the type int or float")
```

You didn't need to make the `layout`, but in doing so you would have lost some ways that `behavior` could pass through to the final output and you weren't handling the `args` and `kwargs` safely: a user could pass the `data` as a keyword; it's not a position-only argument.

The `awkward.type` function works on `akarray` as well as `layout`, but when we guarantee that it's an `ak.Array`, we can even use its `type` property, no special function needed.

src/vector/backends/numpy.py
```diff
@@ -972,6 +981,11 @@ def __array_finalize__(self, obj: typing.Any) -> None:  # type: ignore[override]
         if obj is None:
             return
 
+        if not _is_type_safe(numpy.array(self.tolist())):
+            raise TypeError(
+                "a coordinate must be of the type numpy.integer or numpy.floating"
+            )
```
jpivarski, line 987: This (and following) is still doing an expensive conversion of a NumPy array to a Python list (and back again) to check for type safety. Anywhere that you have `tolist()` is a red flag.

src/vector/backends/object.py
```diff
@@ -3091,6 +3087,11 @@ def obj(**coordinates: float) -> VectorObject:
     """
     is_momentum = False
     generic_coordinates = {}
+
+    for _, value in coordinates.items():
+        if not isinstance(value, (int, float)) or isinstance(value, bool):
+            raise TypeError("a coordinate must be of the type int or float")
```
jpivarski, line 3093:
```suggestion
        if not isinstance(value, numbers.Real) or isinstance(value, bool):
            raise TypeError("a coordinate must be of the type int or float")
```

and `import numbers` at the top of the file. ([See docs](https://docs.python.org/3/library/numbers.html).)

jpivarski (changes requested):
There are still a few instances of expensive data conversions for the sake of checking a data type.

src/vector/backends/numpy.py
```diff
@@ -301,6 +301,26 @@ def _shape_of(
...
+        "CoordinatesNumpy",
+    ],
+) -> bool:
+    for i in range(0, len(array.dtype)):  # type: ignore[arg-type]
+        if not issubclass(
+            array.dtype[i].type, (numpy.integer, numpy.floating)
+        ) or issubclass(array.dtype[i].type, numpy.timedelta64):
+            return False
+
+    return True
```
Saransh-cpp, line 321: The function now takes in a `vector` and checks the `dtype` provided by the user; the cases where the provided `dtype` mismatches with the array entries are handled by `NumPy`.

As a user can only pass in four tuples inside a `dtype` at max, I feel that iterating over them should be fine. Additionally, the `dtype` checks in every `__array_finalize__` method ensure that the `dtype` reaching here does not have a length of 0.

Saransh-cpp (commented):
Thank you for the review, @jpivarski! I have updated the PR with your suggestions! I have also updated the `_is_type_safe` function in `backends/numpy.py` to use the constructed vector. Now the function does not require converting a `NumPy` vector to a list and then to a `NumPy` array.

src/vector/backends/numpy.py
jpivarski: Iterating over metadata is definitely fine! Datasets generally break down into small-metadata and big-data, where the metadata provides a description of the data. Since it's by definition small, it's always okay to iterate over metadata, whereas iterating over data is a thing to do carefully (with NumPy, compiled code, sparingly...).

jpivarski (approved):
This does it! The type-checks are fully general (no missing cases that I can see) and only metadata are iterated over, so I don't see any potential performance issues.

I'd say go ahead and merge it.

Saransh-cpp:
Thank you for the constructive reviews! Merging this!

Issue #209, opened 2022-06-23
Improve the `__repr__` methods of Awkward classes


Saransh-cpp:
The awkward classes representing the coordinates have the default reprs that can be improved.

> I wonder if these reprs could be improved a bit?

_Originally posted by @henryiii in https://github.com/scikit-hep/vector/pull/207#discussion_r903052328_

PR #211, opened 2022-06-24
Implement doctests in CI

### LINKED ISSUE #208: Run doctests
Run the current examples as doctest to ensure they do not go outdated.

> Also is xdoctest or something like that being used to check the examples? It seems a bit easy for them to get out of date. I'm usually not a big fan of doc testing, but vector is simple enough it might not be that difficult to set up.

_Originally posted by @henryiii in https://github.com/scikit-hep/vector/pull/207#discussion_r903052328_

Saransh-cpp:
Closes #208 

- Add `xdoctests` to `test` dependepncies
- Fix the already written doctests and make them deterministic
- Run doctests in the CI

### DIFF
--- a/src/vector/backends/awkward.py
+++ b/src/vector/backends/awkward.py
@@ -108,6 +108,7 @@ class AzimuthalAwkward(CoordinatesAwkward, Azimuthal):
         the fields in ``array``.
 
         Examples:
+            >>> import vector
             >>> import awkward as ak
             >>> a = ak.Array([{"x": [1, 2]}, {"y": [1]}])
             >>> az = vector.backends.awkward.AzimuthalAwkward.from_fields(a)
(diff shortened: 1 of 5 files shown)


src/vector/backends/awkward.py
```diff
@@ -1113,11 +1130,10 @@ def longitudinal(self) -> LongitudinalAwkward:
         to access the coordinates.
 
         Examples:
+            >>> import vector
             >>> a = vector.Array(
             ...       [{"px": 1, "py": 1.1, "z": 0.1, "m": 1}, {"px": 2, "py": 2.2, "z": 0.2, "m": 3}],
             ... )
-            >>> a.longitudinal
-            <vector.backends.awkward.LongitudinalAwkwardZ object at 0x0000028231215A80>
```
henryiii, line 1120: I thought xdoctest recognized the pointers here in an object default repr and still compared correctly? (not sure it's important to show, but just thinking it should have worked if that's the problem)

Saransh-cpp: Yes, locally xdoctest fails the doctests on these outputs. I will get them up in the PR closing #209!

PR #212, opened 2022-06-25
Add custom reprs to awkward coordinate classes

### LINKED ISSUE #209: Improve the `__repr__` methods of Awkward classes

Saransh-cpp:
Closes #209 

- Add custom `__repr__` methods
- Update the doctests to reflect the same

### DIFF
--- a/src/vector/backends/awkward.py
+++ b/src/vector/backends/awkward.py
@@ -100,6 +100,9 @@ class AzimuthalAwkward(CoordinatesAwkward, Azimuthal):
     to construct azimuthal type objects.
     """
 
+    def __repr__(self) -> str:
+        return type(self).__name__ + str(self.elements)
+
     @classmethod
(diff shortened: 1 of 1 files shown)


src/vector/backends/awkward.py
```diff
@@ -100,6 +100,9 @@ class AzimuthalAwkward(CoordinatesAwkward, Azimuthal):
     to construct azimuthal type objects.
     """
 
+    def __repr__(self) -> str:
+        return type(self).__name__ + str(self.elements)
```
henryiii, line 104:
```suggestion
        return f"{type(self).__name__}{self.elements}"
```

henryiii (approved):
Switch to f-strings, slightly faster and cleaner, otherwise good!

PR #223, opened 2022-07-05
chore: migrate to hatchling

Saransh-cpp:
- add `.flake8` for `flake8` config
- remove `setup-cfg-fmt` pre-commit
- remove `setup.*`
- migrate to hatchling


pyproject.toml
```diff
@@ -1,48 +1,179 @@
 [build-system]
 requires = [
-    "setuptools>=42.0",
-    "setuptools_scm[toml]>=3.4"
+    "hatchling>=1.4.1",
```
henryiii, line 3: Why 1.4.1? I think 1.0 dropped 3.6 when 2.7 was dropped.

pyproject.toml
```diff
@@ -1,48 +1,179 @@
...
+dynamic = [
+    "version",
+]
+
+[[project.authors]]
+name = "Jim Pivarski, Henry Schreiner, Eduardo Rodrigues"
+email = "eduardo.rodrigues@cern.ch"
+
+[project.optional-dependencies]
+all = [
```
henryiii, line 53: I'm beginning to recommend dropping `all`, it doesn't work well with static config and isn't really useful, IMO; you really shouldn't be installing docs and test and style and everything else all in one venv, increases chances of conflicts.

pyproject.toml
```diff
@@ -1,48 +1,179 @@
...
+Documentation = "https://vector.readthedocs.io/"
+Homepage = "https://github.com/scikit-hep/vector"
+
+[tool.hatch]
+version.source = "vcs"
+build.hooks.vcs.version-file = "src/vector/version.py"
+
+[tool.hatch.build.targets.sdist]
+include = [
+    "/src",
```
henryiii, line 115: Is this needed? It should be automatic, as long as the `.gitignore` is accurate, I think?

.pre-commit-config.yaml
```diff
@@ -34,11 +34,6 @@ repos:
     hooks:
       - id: isort
 
-  - repo: https://github.com/asottile/setup-cfg-fmt
-    rev: v1.20.1
-    hooks:
-      - id: setup-cfg-fmt
-
   - repo: https://github.com/mgedmin/check-manifest
```
henryiii, line 37: This can probably be dropped too, I think it's (last I checked) not usable with non-setuptools projects (and isn't as important since there's no MANIFEST.in, it uses .gitignore instead)

pyproject.toml
henryiii: I've generally been leaving the min version off. There aren't too many old and broken versions laying around, like setuptools.

Saransh-cpp: This was autogenerated. Yes, I think the tests are failing for python 3.6 because of this. Thanks!

Saransh-cpp:
Thank you for the review, @henryiii! Weirdly, the disassemble check fails due to no `test_extras` dependencies being installed -
```
WARNING: vector 0.1.dev1+g5255483 does not provide the extra 'test_extras'
```

Saransh-cpp:
Turns out, `pyproject.toml` has an issue with having an `_` in the `extra`s name.

Saransh-cpp:
Works with `-`! Thanks, @henryiii!

Saransh-cpp:
Should I go ahead and merge this?

PR #230, opened 2022-07-27
chore: add `pyproject-fmt` pre-commit hook

Saransh-cpp:
Now that `vector` uses only `pyproject.toml` for specifying all the build and packaging-related things, I think adding this hook would be a good idea.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -34,6 +34,11 @@ repos:
     hooks:
       - id: isort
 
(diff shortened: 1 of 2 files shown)


Saransh-cpp:
I think this should be okay to merge without a review as `pyproject-fmt` is in the process of being added to `Scikit-HEP` developer pages - https://github.com/scikit-hep/scikit-hep.github.io/pull/214!

PR #233, opened 2022-08-01
docs: add a developer guide

### LINKED ISSUE #196: Add a developer guide specific to `Vector`?
I was wondering if we should add a developer guide specific to vector and link the [Scikit-HEP  developer information](https://scikit-hep.org/developer) within it.

The basic structure of the `.rst` file might look something like this -
- Installing vector
- Activating pre-commit
- Testing vector
- Documenting vector
- The CI/CD pipeline of vector

Should I proceed with this, or should we directly link the Scikit-HEP guide in the documentation?

Saransh-cpp:
Closes #196


henryiii:
I'm not sure about the move. `.github/CONTRIBUTING.md` is the canonical place for this (or `CONTRIBUTING.md`); GitHub will even suggest a user read this when starting a pull request. Does it really need to be rendered into the documentation? "Normal" readers don't need this info, only users who also are capable of looking at GitHub repo files.

Saransh-cpp:
> I'm not sure about the move. `.github/CONTRIBUTING.md` is the canonical place for this (or `CONTRIBUTING.md`); GitHub will even suggest a user read this when starting a pull request. Does it really need to be rendered into the documentation? "Normal" readers don't need this info, only users who also are capable of looking at GitHub repo files.

This makes sense. I wanted to have a section similar to this - https://awkward-array.org/quickstart.html#for-developers - in `vector`'s documentation. I think I can add something like this -

## Contributing to Vector
If you would like to contribute to vector, please have a look at our Contributing Guidelines.

and redirect to the `.github/CONTRIBUTING.md` file.

henryiii:
Thanks! Sorry I didn't get to this sooner. It's perfect timing for our IRIS-HEP retreat, though!

PR #243, opened 2022-08-17
docs: Add CITATION.cff Citation File Format file

### LINKED ISSUE #238: Add `CITATION.cff` file
`Vector` recently had its first citation - https://arxiv.org/abs/2205.09137 (citation [138]) - which funnily mentions "T. G. Badger" as an author/co-author/contributor. "T. G. Badger" is [The Gitter Badger](https://github.com/gitter-badger), the Gitter bot.

I think now would be a good time to add a `CITATION` file to `Vector`.

This would require some discussion with the maintainers.

Reference - https://github.com/scikit-hep/scikit-hep.github.io/pull/254
cc: @amangoel185

matthewfeickert:
Resolves #238

Add Citation File Format file to repo to get repository cite button on GitHub and citation support on Zenodo.
   - c.f. https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/creating-a-repository-on-github/about-citation-files

The schema used is validated by [`cffconvert`](https://pypi.org/project/cffconvert/)

```console
$ pipx install cffconvert
$ cffconvert --validate --infile CITATION.cff 
Citation metadata are valid according to schema version 1.2.0.
```

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,0 +1,26 @@
+cff-version: 1.2.0
+message: "Please cite the following works when using this software."
+type: software
+title: "vector"
+abstract: "Vector is a Python library for 2D, 3D, and Lorentz vectors, especially arrays of vectors, to solve common physics problems in a NumPy-like way."
+authors:
+- family-names: "Schreiner"
+  given-names: "Henry"
+  orcid: "https://orcid.org/0000-0002-7833-783X"
+  affiliation: "Princeton University"
+- family-names: "Pivarski"
+  given-names: "Jim"
+  orcid: "https://orcid.org/0000-0002-6649-343X"
+  affiliation: "Princeton University"
+- family-names: "Chopra"
+  given-names: "Saransh"
+  orcid: "https://orcid.org/0000-0003-3046-7675"
+  affiliation: "University of Delhi"
+doi: 10.5281/zenodo.5942082
+repository-code: "https://github.com/scikit-hep/vector"
+url: "https://vector.readthedocs.io/"
+keywords:
+  - python
+  - vectors
+  - scikit-hep
+license: "BSD-3-Clause"


CITATION.cff
```diff
@@ -0,0 +1,22 @@
...
+abstract: "Vector is a Python library for 2D, 3D, and Lorentz vectors, especially arrays of vectors, to solve common physics problems in a NumPy-like way."
+authors:
+- family-names: "Schreiner"
+  given-names: "Henry"
+  orcid: "https://orcid.org/0000-0002-7833-783X"
+  affiliation: "Princeton University"
+- family-names: "Pivarski"
+  given-names: "Jim"
+  orcid: "https://orcid.org/0000-0002-6649-343X"
+  affiliation: "Princeton University"
```
matthewfeickert, line 14: @henryiii @jpivarski please check your author information and let me know if there should be other authors added.

matthewfeickert:
In https://github.com/scikit-hep/hist/pull/416 the `CITATION.cff` had to be excluded from the `MANIFEST.in`, but I don't know if anything like that needs to be done with `hatchling`.

matthewfeickert:
Small reminder ping to @henryiii @jpivarski.

CITATION.cff
jpivarski: Given how much he's added to it, I think @Saransh-cpp (Saransh Chopra) should definitely be an author. I don't think he has an [ORCID](https://orcid.org/) yet.

Saranch: It's easy to get one. This is a unique ID that follows all of your academic publications (including code) to disambiguate papers written by people with the same names.

jpivarski: These sum up the history rather well (committer # 2 is pre-commit-ci[bot]).

image

image

image

CITATION.cff
```diff
@@ -0,0 +1,26 @@
...
+- family-names: "Chopra"
+  given-names: "Saransh"
+  orcid: "https://orcid.org/fix-me"
```
matthewfeickert, line 17: @Saransh-cpp once you have a ORCID can you please replace the `fix-me` with it here as a suggested edit?

CITATION.cff
Saransh-cpp: Thanks, @jpivarski! I haven't used ORCID ID much, but I believe I had one. I will add it below!

CITATION.cff
Saransh-cpp: Thanks for taking this up, @matthewfeickert! Here is my ORCID ID -
```suggestion
  orcid: "https://orcid.org/0000-0003-3046-7675"
```

henryiii:
> the CITATION.cff had to be excluded from the MANIFEST.in

It should be in the SDist by default but not the wheel. Don't see harm in having it in the SDist, though it could be manually excluded if we want to. It's tool.hatch.sdist.exclude or something of that sort.

Saransh-cpp:
> It's tool.hatch.sdist.exclude or something of that sort.

```toml
[tool.hatch.build.targets.sdist]
exclude = [
  "./CITATION.cff"
]
```
should work!

https://hatch.pypa.io/latest/build/#configuration

matthewfeickert:
While

```toml
[tool.hatch.build.targets.sdist]
exclude = [
  "/CITATION.cff"
]
```

does work to ignore it from the sdist

```console
$ python -m build . && python -m tarfile --list dist/*.tar.gz | grep "CITATION.cff"
$ echo $?
1
```

I see that `vector` isn't ignoring any other files in the repo, and so for a single text file I don't think it is worth it to ignore the `CITATION.cff`. So I'll just leave this PR as adding the `CITATION.cff` only. 

This should be ready for review from the 3 of you now.

Saransh-cpp (approved):

henryiii (approved):

matthewfeickert:
As 2 of the authors have approved this I'm going to merge this.

PR #244, opened 2022-08-18
feat: add git archive support

Saransh-cpp:
Updates `vector` with the new versioning feature (or bug fix?) mentioned here - https://github.com/scikit-hep/scikit-hep.github.io/pull/250


henryiii:
I'd call it a feature. `git archive` (including the autogenerated ones from GitHub) will have this file rendered out so you can better track where it came from. And setuptools_scm 7+ will pick this up for the version, as well.

PR #251, opened 2022-09-05
chore: remove Python `3.6` support

### LINKED ISSUE #250: Drop Python `3.6` support
Just saw https://github.com/scikit-hep/awkward/pull/1661

I remember @henryiii mentioning dropping Python `3.6` support. Maybe this should be dropped in Vector `v1.0`?

Saransh-cpp:
Fixes #250

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 from pathlib import Path
 
 import nox
--- a/noxfile.py
+++ b/noxfile.py
@@ -2,7 +2,7 @@ from pathlib import Path
 
 import nox
 
-ALL_PYTHONS = ["3.6", "3.7", "3.8", "3.9", "3.10", "3.11"]
+ALL_PYTHONS = ["3.7", "3.8", "3.9", "3.10", "3.11"]
 
 nox.options.sessions = ["lint", "tests", "doctests"]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,7 +15,7 @@ keywords = [
 license = "BSD-3-Clause"
 maintainers = [ {name = "The Scikit-HEP admins", email = "scikit-hep-admins@googlegroups.com"} ]
 authors = [ {name = "Jim Pivarski, Henry Schreiner, Eduardo Rodrigues", email = "eduardo.rodrigues@cern.ch"} ]
-requires-python = ">=3.6"
+requires-python = ">=3.7"
 dependencies = [
   'importlib-metadata>=0.22; python_version < "3.8"',
   "numpy>=1.13.3",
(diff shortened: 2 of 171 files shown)


pyproject.toml
```diff
@@ -51,7 +50,7 @@ awkward = [
 ]
 dev = [
   "awkward>=1.2",
-  'numba>=0.50; python_version >= "3.6"',
+  'numba>=0.50; python_version >= "3.7"',
```
henryiii, line 53: This is always true?

noxfile.py
```diff
@@ -2,7 +2,7 @@
 
 import nox
 
-ALL_PYTHONS = ["3.6", "3.7", "3.8", "3.9", "3.10"]
+ALL_PYTHONS = ["3.7", "3.8", "3.9", "3.10"]
```
henryiii, line 5: Can we add 3.11 now, too? And 3.11-dev to the testing matrix? Awkward supports 3.11 now. Some of the dev dependencies will need to require <3.11.

henryiii:
We can change pyupgrade to `--py37plus`, and add the annotations addition to isort (see SKHEP dev pages).

noxfile.py
henryiii: (Actually the 3.11 stuff can be added before 0.9 & this PR)

src/vector/_methods.py
```diff
@@ -194,140 +196,140 @@ def to_Vector4D(self) -> "VectorProtocolLorentz":
         """
         raise AssertionError
 
-    def to_xy(self) -> "VectorProtocolPlanar":
```
Saransh-cpp, line 197: I am not sure how all the quotes from the type hints went away. Was this a pre-commit hook?

jpivarski:
You probably have access to change this (try checking): go to Settings (gear icon), then Branches (left menu), then Set Protection Rules. Inside that, there's a start-typing-then-dropdown-menu for adding new requirements and a way to "X" out old requirements.

image

I did it already for this one (dropped Python 3.6 and added 3.10 and 3.11-dev).

Saransh-cpp:
I don't have access to the Settings tab 😅 Thank you!

src/vector/_methods.py
henryiii: Yes, PyUpgrade will do this if `from __future__ import annotations` is present.

henryiii:
You might have permissions now. Maybe.

henryiii (approved):

pyproject.toml
```diff
@@ -15,7 +15,7 @@ keywords = [
 license = "BSD-3-Clause"
 maintainers = [ {name = "The Scikit-HEP admins", email = "scikit-hep-admins@googlegroups.com"} ]
 authors = [ {name = "Jim Pivarski, Henry Schreiner, Eduardo Rodrigues", email = "eduardo.rodrigues@cern.ch"} ]
-requires-python = ">=3.6"
+requires-python = ">=3.7"
 dependencies = [
   'importlib-metadata>=0.22; python_version < "3.8"',
   "numpy>=1.13.3",
```
henryiii, line 21: I think this is now higher; the first version to support 3.7 was newer than 1.13.3.

henryiii: It's `1.14.5`.

Saransh-cpp:
> You might have permissions now. Maybe.

Yes! Thank you!

PR #254, opened 2022-09-11
fix: `vector.arr` should construct `NumPy` vectors

Saransh-cpp:
Currently, all the examples located in `vector`, as well as the introduction to constructors here - https://github.com/scikit-hep/vector#constructing-a-vector-or-an-array-of-vectors - point to the fact that `vector.arr` should construct `NumPy` vectors. This also seems logical (`Array` -> `awk` , `array` -> `arr`), and is most probably not wrong in the examples.


Saransh-cpp:
Merging this in! (a small bug fix from developer's perspective, but might break scripts that are using this bug as a feature)

PR #259, opened 2022-09-22
fix: fix breakages for Awkward v2

agoose77:
Contrary to what I said in the meeting, we needed a few additional fixes (that I'd already made and forgotten about).


agoose77:
Actually, let me deal with one thing at a time here! :sweat_smile:

PR #278, opened 2022-10-19
feat: update `to_Vector3D` to pass new coordinate values

Naman-Priyadarshi:
## Description

## Checklist

- [ ] Have you checked to ensure there aren't any other open Pull Request for the required change ?
- [ ] Does your submission pass doctests? (`$ xdoctest ./src/vector` or `$ nox -s doctests`)

## Before Merging

- [ ] Summarize commit messages into a brief review of the Pull request.

### DIFF
--- a/src/vector/_methods.py
+++ b/src/vector/_methods.py
@@ -2969,16 +2969,16 @@ class Vector2D(Vector, VectorProtocolPlanar):
         if sum(x is not None for x in (z, theta, eta)) > 1:
             raise TypeError("Only one non-None parameter allowed")
 
-        z = 0
+        coord_value = 0.0
         l_type = LongitudinalZ
         if z is not None:
             coord_value = z
         elif eta is not None:
             coord_value = eta
-            l_type = LongitudinalEta
+            l_type = LongitudinalEta  # type: ignore[assignment]
         elif theta is not None:
             coord_value = theta
-            l_type = LongitudinalTheta
+            l_type = LongitudinalTheta  # type: ignore[assignment]
 
         return self._wrap_result(
             type(self),
--- a/src/vector/_methods.py
+++ b/src/vector/_methods.py
@@ -2969,15 +2969,16 @@ class Vector2D(Vector, VectorProtocolPlanar):
         if sum(x is not None for x in (z, theta, eta)) > 1:
             raise TypeError("Only one non-None parameter allowed")
 
+        z = 0
+        l_type = LongitudinalZ
         if z is not None:
             coord_value = z
-            l_type = LongitudinalZ
         elif eta is not None:
             coord_value = eta
-            l_type = LongitudinalEta  # type: ignore[assignment]
+            l_type = LongitudinalEta
         elif theta is not None:
             coord_value = theta
-            l_type = LongitudinalTheta  # type: ignore[assignment]
+            l_type = LongitudinalTheta
 
         return self._wrap_result(
             type(self),
--- a/src/vector/_methods.py
+++ b/src/vector/_methods.py
@@ -2969,22 +2969,20 @@ class Vector2D(Vector, VectorProtocolPlanar):
         if sum(x is not None for x in (z, theta, eta)) > 1:
             raise TypeError("Only one non-None parameter allowed")
 
-        coord_value = 0.0
-        l_type: type[Longitudinal] = LongitudinalZ
-
         if z is not None:
             coord_value = z
+            l_type = LongitudinalZ
         elif eta is not None:
             coord_value = eta
-            l_type = LongitudinalEta
+            l_type = LongitudinalEta  # type: ignore[assignment]
         elif theta is not None:
             coord_value = theta
-            l_type = LongitudinalTheta
+            l_type = LongitudinalTheta  # type: ignore[assignment]
 
         return self._wrap_result(
             type(self),
-            (*self.azimuthal.elements, 0),
-            [_aztype(self), LongitudinalZ, None],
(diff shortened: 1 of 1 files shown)


Naman-Priyadarshi:
Making changes to test_conversions.py now

src/vector/_methods.py
```diff
@@ -1798,11 +1799,23 @@ class Vector2D(Vector, VectorProtocolPlanar):
     def to_Vector2D(self) -> VectorProtocolPlanar:
         return self
 
-    def to_Vector3D(self) -> VectorProtocolSpatial:
+    def to_Vector3D(self, **coord) -> VectorProtocolSpatial:
+        if len(coord) > 1:
+            raise ArgumentError(coord, "can be `z`, `theta`, or `eta`")
```
henryiii, line 1804: This should not be an argparse error. It probably should be TypeError. But better yet, instead of manually parsing a `**`, it should list the three keywords = None then verify there are at least two None's.

Saransh-cpp:
This is how it works right now -
```py
In [1]: import vector

In [2]: v = vector.obj(x=1, y=2)

In [3]: v.to_Vector3D(z=1)
Out[3]: VectorObject3D(x=1, y=2, z=1)

In [4]: v = vector.obj(px=1, y=2)

In [5]: v.to_Vector3D(z=1)
Out[5]: vector.MomentumObject3D(px=1, py=2, pz=1)
```

Should we also have the option to enter the momentum counterparts of coordinates? Or should we leave it as it is?

jpivarski (approved):
This looks like a good thing to add, except that if it's on `Vector2D.to_Vector3D`, users will expect it to be on 2D → 4D and 3D → 4D as well. @Saransh-cpp, how much effort do you think it would be to add them?

Also, a new feature like this should have tests. Like, one example for each of `z`, `theta`, and `eta`.

src/vector/_methods.py
```diff
@@ -2969,16 +2969,16 @@ def to_Vector3D(
         if sum(x is not None for x in (z, theta, eta)) > 1:
             raise TypeError("Only one non-None parameter allowed")
 
-        z = 0
+        coord_value = 0.0
         l_type = LongitudinalZ
```
henryiii, line 2973: You can set this to something like `l_type: type[Longitudinal] = LongitudinalZ` or something like that, IIRC, to avoid the type ignore.

henryiii:
> users will expect it to be on 2D → 4D and 3D → 4D as well

I was thinking that as well. Though maybe asking people to do 2D -> 3D -> 4D would be easier than adding it to the 2D -> 4D one. But 3D -> 4D does clearly make sense.

jpivarski:
2D → 4D could be implemented in terms of 2D → 3D → 4D.

Also, does this only apply to the single-object backend (`vector.obj`), or does it generalize to all backends? Setting `coord_value` to a scalar `0` might work on array-based backends if it gets broadcasted.

Also, the distinction between `0` and `0.0` is relevant: coordinates are allowed to be any numeric type, including integers (though that's less often desired).

src/vector/_methods.py
Saransh-cpp: I did try this (in fact, IIRC, this was added by you in one of the commits above), but `mypy` still required the type ignore statements :(

Saransh-cpp:
Thanks for the review, @jpivarski and @henryiii! Could you also please take a look at #317 and let me know if this should go in `v1.0.0`?

---

> users will expect it to be on 2D → 4D and 3D → 4D as well. @Saransh-cpp, how much effort do you think it would be to add them?
>
> Also, a new feature like this should have tests. Like, one example for each of z, theta, and eta.

Should be simple to add. I can maybe merge this PR and add the tests + make it available for 2D → 4D (2D → 3D → 4D) and 3D → 4D in another PR.

> Also, does this only apply to the single-object backend (vector.obj), or does it generalize to all backends? Setting coord_value to a scalar 0 might work on array-based backends if it gets broadcasted.

Generalizes to all the backends!

```py
In [1]: import vector

In [2]: v = vector.array(
   ...:     [(1.1, 2.1), (1.2, 2.2), (1.3, 2.3), (1.4, 2.4), (1.5, 2.5)],
   ...:     dtype=[("x", float), ("y", float)],
   ...: )

In [3]: v.to_Vector3D(z=1)
Out[3]: 
VectorNumpy3D([(1.1, 2.1, 1.), (1.2, 2.2, 1.), (1.3, 2.3, 1.), (1.4, 2.4, 1.),
               (1.5, 2.5, 1.)], dtype=[('x', '<f8'), ('y', '<f8'), ('z', '<f8')])

In [4]: v = vector.Array(
   ...:     [
   ...:         [{"x": 1, "y": 1.1}, {"x": 2, "y": 2.2}],
   ...:         [],
   ...:         [{"x": 3, "y": 3.3}]
   ...:     ]
   ...: )

In [5]: v.to_Vector3D(z=1)
Out[5]: <VectorArray3D [[{x: 1, y: 1.1, z: 1, ... z: 1}]] type='3 * var * Vector3D["x": ...'>
```

henryiii:
I think this would be nice for a v1.

How much work is it to keep Awkward v1? If not much, then I'd say we don't need to rush to drop it, it will make adoption of Vector v1 easier. But fine to drop if it is.

Saransh-cpp:
Thanks! Will merge this, create a new PR, and set up stuff for v1!

Saransh-cpp:
Thanks for taking this up, @Naman-Priyadarshi!

PR #282, opened 2022-10-25
chore: use Python 3.11!

Saransh-cpp:
## Description

I am hoping this will work.

## Checklist

- [ ] Does your submission passes doctests? (`$ xdoctest ./src/vector` or `$ nox -s doctests`)

## Before Merging

- [ ] Summarize commit messages into a brief review of the Pull request.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,7 +39,7 @@ jobs:
           - "3.8"
           - "3.9"
           - "3.10"
-          - "3.11-dev"
+          - "3.11"
     name: Check Python ${{ matrix.python-version }}
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -94,6 +94,12 @@ jobs:
       - name: Test package
         run: python -m pytest -ra -m dis --ignore tests/test_notebooks.py
 
+  pass:
+    needs: [pre-commit, checks, discheck]
+    runs-on: ubuntu-latest
+    steps:
+      - run: echo "All jobs passed"
+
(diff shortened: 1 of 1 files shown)


henryiii:
It should, 3.11 for actions should have been released this morning. But we'll see! 🤞

henryiii:
We should add a "pass" check, then only require that one. Then it's much easier to edit required checks for passing. See henryiii/scikit-build-core for an example.

Saransh-cpp:
> I think this only worked because we don't have a macOS check - 3.11 macOS doesn't seem to be there yet. But doesn't affect us. :)

Nice!!

> We should add a "pass" check that depends on all the other ones, then only require that one. Then it's much easier to edit required checks for passing. See henryiii/scikit-build-core for an example.

Added! I don't seem to have the appropriate rights for setting a job as "required". Jim did grant me the rights for accessing the settings tab of vector, but I cannot see any option to set (or un-set) a job as "required". Could you please make the required changes?

jpivarski:
**In settings, do you see the following?**

image

**That should take you to**

image

**"Edit" should go to**

image

**And then if you start typing in that box, you should get**

image

@Saransh-cpp, your permissions level is "Maintain," and there is one higher, "Admin." If you can't do the above and need to, then that's how it can be fixed.

Saransh-cpp:
My settings page looks like this -

image

I am fine with being at the "Maintain" level. I don't think we would need any more "required" changes after this one (thanks to henry's suggestion!) 😅 

Thank you for updating the requirements! Though as Henry suggested, it would be better to keep only the "pass" check as "required" to avoid updating these requirements every time a new job/workflow is added.

PR #305, opened 2022-12-27
fix: update discheck

Saransh-cpp:
## Description

## Checklist

## Before Merging

### DIFF
--- a/tests/test_compute_features.py
+++ b/tests/test_compute_features.py
@@ -215,7 +215,7 @@ def analyze_return(node, context):
 
     assert node[0].kind == "return", "compute function must end with a 'return'"
     assert len(node[0]) == 2
-    assert node[0][0].kind == "return_expr"
+    assert node[0][0].kind in ("ret_expr", "return_expr")
     assert len(node[0][0]) == 1
     expr(node[0][0][0])
     assert node[0][1].kind == "RETURN_VALUE"
@@ -277,8 +277,12 @@ def analyze_expression(node, context):
         analyze_callable(expr(node[0]), context)
 
         for argument in node[1:-1]:
-            assert argument.kind == "pos_arg", "only positional arguments"
-            analyze_expression(expr(argument[0]), context)
+            if argument.kind == "pos_arg":
+                expr_arg = argument[0]
+            else:
+                expr_arg = argument
+            assert expr_arg.kind == "expr", "only positional arguments"
+            analyze_expression(expr(expr_arg), context)
 
     elif is_nan_to_num(node):
         analyze_expression(expr(node[1]), context)
--- a/tests/test_compute_features.py
+++ b/tests/test_compute_features.py
@@ -215,7 +215,7 @@ def analyze_return(node, context):
 
     assert node[0].kind == "return", "compute function must end with a 'return'"
     assert len(node[0]) == 2
-    assert node[0][0].kind == "ret_expr"
+    assert node[0][0].kind == "return_expr"
     assert len(node[0][0]) == 1
     expr(node[0][0][0])
     assert node[0][1].kind == "RETURN_VALUE"


Saransh-cpp:
Thank you so much, @jpivarski! :)

jpivarski:
Uncompyle6 just made a new release: [3.9.0](https://github.com/rocky/python-uncompyle6/releases/tag/3.9.0) (December 22, 2022). The most recent version before that was more than a year ago: [3.8.0](https://github.com/rocky/python-uncompyle6/releases/tag/3.8.0) (October 29, 2021).

This release contained two trivial changes to the way bytecode is presented (without any changes to the bytecode itself because it's all Python 3.8, anyway).

You found the first one: https://github.com/rocky/python-uncompyle6/pull/381 changed the spelling of `"ret_expr"` to `"return_expr"`.

The second one changed the structure of bytecode parsing trees, but only by removing a wrapping layer: `"pos_arg" → "expr" → ...` became just `"expr" → ...`. The commit said that it's addressing PyPy support: https://github.com/rocky/python-uncompyle6/commit/2ed211e0d4a5b53d3f373ca05afd4655bf48b192 but the change came in the non-PyPy `else` clause:

```diff
- # number of apply equiv arguments:
- nak = (len(opname_base) - len("CALL_METHOD")) // 3
- rule = (
-     "call ::= expr "
-     + ("pos_arg " * args_pos)
-     + ("kwarg " * args_kw)
-     + "expr " * nak
-     + opname
- )
+ if opname == "CALL_METHOD_KW":
+     args_kw = token.attr
+     rules_str = """
+          expr ::= call_kw_pypy37
+          pypy_kw_keys ::= LOAD_CONST
+     """
+     self.add_unique_doc_rules(rules_str, customize)
+     rule = (
+         "call_kw_pypy37 ::= expr "
+         + ("expr " * args_kw)
+         + " pypy_kw_keys "
+         + opname
+     )
+ else:
+     args_pos, args_kw = self.get_pos_kw(token)
+     # number of apply equiv arguments:
+     nak = (len(opname_base) - len("CALL_METHOD")) // 3
+     rule = (
+         "call ::= expr "
+         + ("expr " * args_pos)
+         + ("kwarg " * args_kw)
+         + "expr " * nak
+         + opname
+     )
```

(See that the `("pos_arg " * args_pos)` has become `("expr " * args_pos)`.) This change was probably unintentional, but all that we care about is that we can distinguish between positional arguments and keyword arguments, and that distinction is preserved.

For a sample function (`vector._compute.planar.add.xy_xy`), the difference that this makes is

```diff
<                  1. pos_arg
<                     expr
<                                        12  LOAD_FAST                'lib'
<                  2. pos_arg
<                     expr
<                                        14  LOAD_FAST                'rho2'
<                  3. pos_arg
<                     expr
<                                        16  LOAD_FAST                'phi2'
---
>                  1. expr
>                                    12  LOAD_FAST                'lib'
>                  2. expr
>                                    14  LOAD_FAST                'rho2'
>                  3. expr
>                                    16  LOAD_FAST                'phi2'
39,47c36,41
<                  1. pos_arg
<                     expr
<                                        24  LOAD_FAST                'lib'
<                  2. pos_arg
<                     expr
<                                        26  LOAD_FAST                'rho2'
<                  3. pos_arg
<                     expr
<                                        28  LOAD_FAST                'phi2'
---
>                  1. expr
>                                    24  LOAD_FAST                'lib'
>                  2. expr
>                                    26  LOAD_FAST                'rho2'
>                  3. expr
>                                    28  LOAD_FAST                'phi2'
```

(The `expr` is no longer wrapped by a single-argument `pos_arg`.)

The fix that I added makes our test insensitive to these two changes in uncompyle6. Now that they have a new version out, it _might_ become possible to use a later version of Python 3.8 or even a later version of Python, but I haven't tested it. When Python 3.8 gets end-of-lifed, I'll revisit this.

It doesn't look like a lot of new `_compute` functions are being added. This disassemble test is to guard against `_compute` functions being added in such a way that would limit our ability to add backends in the future. If no one is adding `_compute` functions or modifying them, that's not much of a danger.

It took me so long to write this note that the PR is already merged now. I hadn't noticed that auto-merge was on.

rocky:
> Uncompyle6 just made a new release: ...
> You found the first one: [rocky/python-uncompyle6#381](https://github.com/rocky/python-uncompyle6/pull/381) changed the spelling of `"ret_expr"` to `"return_expr"`.

Let me explain. This was to correspond to the term Python's AST uses ["Return"](https://docs.python.org/3/library/ast.html#ast.Return). Note that a decompilation grammar the Python grammar and AST will always be different, but to the extent they can be similar that is probably a good thing. The lower-casing and underscores is intentional to indicate that these two things _are_ different.

It is possible that "return" may get used in the future to conform to Python's terminology better.

> 
> The second one changed the structure of bytecode parsing trees...
> 
> (See that the `("pos_arg " * args_pos)` has become `("expr " * args_pos)`.) This change was probably unintentional,

You are correct. It was unintentional. Thanks for pointing this out. 
> 
> The fix that I added makes our test insensitive to these two changes in uncompyle6.

Although in the current master, the parse tree is now more like the way it was in in 3.8.0, I am not planning on a release of uncompyle6 in the near future.

Saransh-cpp:
Oops, I merged too quickly. Thank you for the amazingly detailed explanations! I don't think any more changes are required here? Unless we want to switch to the master branch of `uncompyle6` that adds back "pos_args"?

rocky:
You all ultimately should decide what's best for you. I will however try to give you input hopefully for you to figure out what is best for you. 

For Python 3.7 and Python 3.8 there are in fact two decompilers `uncompyle6` which you are using and [`decompyle3` ](https://github.com/rocky/python-decompile3) which tries to clean up problems with `uncompyle6` by focusing on those two specific versions. There isn't a PyPI release of that, and there isn't likely to be one. See the paragraph _after_ next for why.

So what happens is changes sometimes get made in one repository and then get ported to the other. The decompyle3 to uncompyle3 direction is often because that tries to be cleaner. The uncompyle6 to decompyle3 direction is often because people use uncompyle6 more - possibly because there is a PyPI package for it and it covers more versions of Python and is older. So bugs are reported there and are sometimes fixed there first. But in truth if you are only interested in Python 3.7 or 3.8 decompyle3 does a better job at decompilation.

If this isn't complicated enough, there is yet _another_ decompiler version which focuses on 3.8 and now 3.9 a little.  In that,  I am using up front more sophisticated control flow detection via https://github.com/rocky/python-control-flow.   The 3.8-3.9 decompiler however is private and is likely to remain that way for a while. 

As I was trying to sync up decompyle3 with the with the changes to uncompyle6, I now have come to understand that possibly the removal of `pos_args` was probably again motivated by a slight desire to match Python's AST closer. Python's AST  does not have a notion of "positional args", and does have a base class called [`ast.expr`](https://docs.python.org/3.11/library/ast.html#ast.AST) . 

In sum, from the decompiler development side, as with so many things, there is no clear winner here. And as I said, it is possible that `return_expr` will get renamed to `return` to better match the AST since there is a desire to try to make this simple and clear so the people might understand this mirky stuff more clearly.

jpivarski:
For @Saransh-cpp: there's no need to reopen or unmerge this PR: it's done. I was just in the process of documenting the explanations of the changes.

For @rocky: thanks for the heads-up about upcoming changes. When they come, we'll adjust. This test asserts on just about everything, so if the kind of `return_expr` becomes `return`, that will be easy to recognize and we'll loosen our test. I like to try to find the change in uncompyle6 to justify the change, though; that's why I reported PR and commit numbers here.

Just to be clear, this project only uses uncompyle6 in a test, since we have a suite of functions that need to remain "simple" (simplicity is defined by the tests) so that we don't prevent ourselves from being able to add new array backends in the future. That test can run in any one version of Python that isn't end-of-lifed. Vector can run in Python versions that uncompyle6 doesn't support. (For Python 3.8, we have until October 14, 2024!)

(I had [another potential project](https://github.com/diana-hep/rejig) that would have used uncompyle6 in a more extensive way, but that one never became a finished product.)

PR #315, opened 2023-02-17
chore: move to using Ruff

henryiii:
- chore: move to using Ruff
- style: make Ruff happy


src/vector/backends/object.py
```diff
@@ -742,31 +734,11 @@ def _wrap_result(
...
-
-        elif (
-            len(returns) == 3
+            or (len(returns) == 3 and returns[2] is None)
             and isinstance(returns[0], type)
             and issubclass(returns[0], Azimuthal)
             and isinstance(returns[1], type)
             and issubclass(returns[1], Longitudinal)
-            and returns[2] is None
         ):
```
henryiii, line 742: I had to manually make these changes, but pretty impressed Ruff noticed it could be made. :)

.pre-commit-config.yaml
```diff
@@ -48,18 +41,9 @@ repos:
...
+          - numpy
           - packaging
 
-  - repo: https://github.com/PyCQA/flake8
-    rev: 6.0.0
-    hooks:
-      - id: flake8
-        additional_dependencies:
-          - flake8-bugbear
-          - flake8-docstrings
```
henryiii, line 60: For some reason, Ruff's "D" code catches a lot of stuff the old one didn't. So I've removed it for now, and we can enable new codes (including "D") in the future.

tests/backends/test_numba_object.py
```diff
@@ -694,7 +694,7 @@ def momentum_rhophizt():
     assert out.pt == pytest.approx(2)
     assert out.phi == pytest.approx(3.3)
     assert out.z == pytest.approx(5)
-    assert out.E == pytest.approx(10)
+    assert out.E == pytest.approx(10)  # noqa: SIM300
```
henryiii, line 697: Yoda conditions thinks ".E" is a constant.

tests/backends/test_numba_object.py
```diff
@@ -1368,11 +1368,11 @@ def get_true(v):
 def test_operator_truth():
     @numba.njit
     def get_true(v):
-        return True if v else False
+        return True if v else False  # noqa: SIM210
 
     @numba.njit
     def get_false(v):
-        return False if v else True
+        return False if v else True  # noqa: SIM211
```
henryiii, line 1375: We don't support `not v`, which is a little strange, but just noqa'd it. `not bool(v)` probably would have worked, since I think `bool(v)` above worked.

PR #319, opened 2023-02-22
feat: allow passing coordinates to to_Vector*D

### LINKED ISSUE #276: Allow users to pass new coordinate values in `to_Vector{X}D` and `to_{new coordinate names}`
### Describe the potential feature

Right now users can convert a 2D vector to 3D or 4D using `to_Vector3D()` and `to_Vector4D()`, but the returned vector always possesses 0 as the value of the new generated coordinate.

Example:
```py
In [1]: import vector

In [2]: v = vector.obj(x=1, y=2)

In [3]: v.to_Vector3D()
Out[3]: VectorObject3D(x=1, y=2, z=0)

In [4]: v.to_Vector3D(z=1)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In [4], line 1
----> 1 v.to_Vector3D(z=1)

TypeError: to_Vector3D() got an unexpected keyword argument 'z'
```

This can be extended such that the users can pass in the value of the newly added coordinate.

Example:
```py
In [1]: import vector

In [2]: v  = vector.obj(x=1, y=2)

In [3]: v
Out[3]: VectorObject2D(x=1, y=2)

In [4]: v.to_Vector3D(z=1)
Out[4]: VectorObject3D(x=1, y=2, z=1)
```

Similarly for 3D and 4D vectors. This can then be further propagated to the `to_{new coordinate name}` (`to_xyzt`, `to_rhophi`, ...) methods.

### Motivation

The discussion on gitter (between @jpivarski and @henryiii). Quoting a part here -

> For 3D and 4D vectors, you probably want a specialized function, since that 4th dimension is qualitatively different from the other 3. Maybe something like scale_spatial(-1). (There's already a scale; it does the same thing as multiplication by a scalar.)
>
> But even without that, it would also be nice if functions like to_Vector4D, which is increasing the dimensionality of the 3D vector (x.to_Vector3D()) in this case, could take parameters like E= or mass= to assign a value to the new dimension. As it is, to_Vector4D applied to a 3D vector creates a 4th dimension whose value is 0 in Cartesian coordinates. To get the original x.E in it, I had to add vectors. If the coordinate were expressed as a mass, the above trick wouldn't work with x.mass because it would add after assigning a mass; non-Cartesian coordinates don't commute.

### Possible Implementation

Changing this -

https://github.com/scikit-hep/vector/blob/93cb78a017707f42b961b3a74e34ffb79c37f85d/src/vector/_methods.py#L1701-L1707

to this -
```py
def to_Vector3D(self, **coord) -> VectorProtocolSpatial:
    if len(coord) > 1:
        raise ArgumentError(coord, "can be `z`, `theta`, or `eta`")

    coord_type, coord_value = list(coord.keys())[0], list(coord.values())[0]

    if coord_type == "z":
        l_type = LongitudinalZ
    elif coord_type == "eta":
        l_type = LongitudinalEta
    elif coord_type == "theta":
        l_type = LongitudinalTheta

    return self._wrap_result(
        type(self),
        self.azimuthal.elements + (coord_value,),
        [_aztype(self), l_type, None],
        1,
    )
```

works -

```py
In [1]: import vector

In [2]: v = vector.obj(x=1, py=2)

In [3]: v.to_Vector3D(z=3)
Out[3]: vector.MomentumObject3D(px=1, py=2, pz=3)

In [4]: v.to_Vector3D(eta=3)
Out[4]: vector.MomentumObject3D(px=1, py=2, eta=3)

In [5]: v.to_Vector3D(theta=3)
Out[5]: vector.MomentumObject3D(px=1, py=2, theta=3)
```

Saransh-cpp:
## Description

Closes #276 

Adds test and docs for #278 

## Checklist

- [X] Does your submission pass the doctests? (`$ xdoctest ./src/vector` or `$ nox -s doctests`)

## Before Merging

### DIFF
--- a/src/vector/_methods.py
+++ b/src/vector/_methods.py
@@ -2962,7 +2962,7 @@ class Vector2D(Vector, VectorProtocolPlanar):
     def to_Vector3D(self) -> VectorProtocolSpatial:
         return self._wrap_result(
             type(self),
-            self.azimuthal.elements + (0,),
+            (*self.azimuthal.elements, 0),
             [_aztype(self), LongitudinalZ, None],
             1,
         )
@@ -2970,7 +2970,7 @@ class Vector2D(Vector, VectorProtocolPlanar):
     def to_Vector4D(self) -> VectorProtocolLorentz:
         return self._wrap_result(
             type(self),
-            self.azimuthal.elements + (0, 0),
+            (*self.azimuthal.elements, 0, 0),
             [_aztype(self), LongitudinalZ, TemporalT],
             1,
         )
@@ -4026,9 +4026,7 @@ def _handler_of(*objects: VectorProtocol) -> VectorProtocol:
     for obj in objects:
         if not isinstance(obj, Vector):
             continue
-        if handler is None:
-            handler = obj
-        elif _get_handler_index(obj) > _get_handler_index(handler):
+        if handler is None or _get_handler_index(obj) > _get_handler_index(handler):
             handler = obj
 
     assert handler is not None
--- a/src/vector/_methods.py
+++ b/src/vector/_methods.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2019-2021, Jonas Eschle, Jim Pivarski, Eduardo Rodrigues, and Henry Schreiner.
+# Copyright (c) 2019-2023, Jonas Eschle, Jim Pivarski, Eduardo Rodrigues, and Henry Schreiner.
 #
 # Distributed under the 3-clause BSD license, see accompanying file LICENSE
 # or https://github.com/scikit-hep/vector for details.
--- a/src/vector/_methods.py
+++ b/src/vector/_methods.py
@@ -2967,31 +2967,63 @@ class Vector2D(Vector, VectorProtocolPlanar):
         eta: float | None = None,
     ) -> VectorProtocolSpatial:
         if sum(x is not None for x in (z, theta, eta)) > 1:
-            raise TypeError("Only one non-None parameter allowed")
+            raise TypeError("Only one non-None longitudinal coordinate allowed")
 
-        coord_value = 0.0
+        l_value = 0.0
         l_type: type[Longitudinal] = LongitudinalZ
         if z is not None:
-            coord_value = z
+            l_value = z
         elif eta is not None:
-            coord_value = eta
+            l_value = eta
             l_type = LongitudinalEta
         elif theta is not None:
-            coord_value = theta
+            l_value = theta
             l_type = LongitudinalTheta
 
         return self._wrap_result(
             type(self),
-            (*self.azimuthal.elements, coord_value),
+            (*self.azimuthal.elements, l_value),
             [_aztype(self), l_type, None],
             1,
         )
 
-    def to_Vector4D(self) -> VectorProtocolLorentz:
+    def to_Vector4D(
+        self,
+        *,
+        z: float | None = None,
+        theta: float | None = None,
+        eta: float | None = None,
+        t: float | None = None,
+        tau: float | None = None,
+    ) -> VectorProtocolLorentz:
+        if sum(x is not None for x in (z, theta, eta)) > 1:
+            raise TypeError("Only one non-None longitudinal coordinate allowed")
+        elif sum(x is not None for x in (t, tau)) > 1:
+            raise TypeError("Only one non-None temporal coordinate allowed")
+
+        t_value = 0.0
+        t_type: type[Temporal] = TemporalT
+        if t is not None:
+            t_value = t
+        elif tau is not None:
+            t_value = tau
+            t_type = TemporalTau
+
+        l_value = 0.0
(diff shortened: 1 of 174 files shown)


src/vector/_methods.py
```diff
@@ -3008,11 +3065,39 @@ def to_Vector2D(self) -> VectorProtocolPlanar:
...
+    def to_Vector4D(
+        self,
+        *,
+        t: float | None = None,
+        tau: float | None = None,
+    ) -> VectorProtocolLorentz:
+        """
+        Converts a 3D vector to 4D vector.
+
+        The scalar temporal coordinate are broadcasted for NumPy and Awkward vectors.
```
Saransh-cpp, line 3077:
```suggestion
        The scalar temporal coordinate is broadcasted for NumPy and Awkward vectors.
```

src/vector/_methods.py
```diff
@@ -2966,32 +2966,97 @@ def to_Vector3D(
...
+            >>> vec = vector.VectorObject2D(x=1, y=2)
+            >>> vec.to_Vector3D(z=1)
+            VectorObject3D(x=1, y=2, z=1)
+            >>> vec = vector.MomentumObject2D(px=1, py=2)
+            >>> vec.to_Vector3D(z=4)
+            MomentumObject3D(px=1, py=2, pz=4)
+        """
         if sum(x is not None for x in (z, theta, eta)) > 1:
-            raise TypeError("Only one non-None parameter allowed")
+            raise TypeError("Only one non-None longitudinal coordinate allowed")
```
jpivarski, line 2986: The above is helpful; maybe this is even more helpful (and it sets the tone for what the other `to_Vector*D` should do).

```suggestion
            raise TypeError("At most one longitudinal coordinate (`z`, `theta`, or `eta`) may be assigned (non-None)")
```

By the way, these `to_Vector*D` functions maintain momentum-ness, right? For example, is it the case that `Momentum2D.to_Vector3D` returns a `Momentum3D`?

If so, then a user might justifiably expect to pass `pz` instead or `z`, but only in the momentum case. Just something to consider, at this point.

When this gets extended to `to_Vector4D`, the time component has a lot of momentum synonyms. The geometric names for the two possible temporal coordinates are `t` and `tau`, but the momentum names are `energy` and `mass`, `E` and `M`, and (I think) `e` and `m`. There was a discussion once about including the lowercase as well as the uppercase.

src/vector/_methods.py
```diff
@@ -2966,32 +2966,97 @@ def to_Vector3D(
...
+            >>> vec.to_Vector4D(z=3, t=4)
+            VectorObject4D(x=1, y=2, z=3, t=4)
+            >>> vec = vector.MomentumObject2D(px=1, py=2)
+            >>> vec.to_Vector4D(z=4, t=4)
+            MomentumObject4D(px=1, py=2, pz=4, E=4)
+        """
+        if sum(x is not None for x in (z, theta, eta)) > 1:
+            raise TypeError("Only one non-None longitudinal coordinate allowed")
+        elif sum(x is not None for x in (t, tau)) > 1:
+            raise TypeError("Only one non-None temporal coordinate allowed")
```
jpivarski, line 3035: Similar error messages: help the user by spelling out what the longitudinal and temporal coordinates are, since the user would more likely be familiar with concrete names like `z`, `theta`, `eta` than "longitudinal".

src/vector/_methods.py
```diff
@@ -3008,11 +3073,43 @@ def to_Vector2D(self) -> VectorProtocolPlanar:
...
+            >>> import vector
+            >>> vec = vector.VectorObject3D(x=1, y=2, z=3)
+            >>> vec.to_Vector4D(t=4)
+            VectorObject4D(x=1, y=2, z=3, t=4)
+            >>> vec = vector.MomentumObject3D(px=1, py=2, pz=3)
+            >>> vec.to_Vector4D(tau=4)
+            MomentumObject4D(px=1, py=2, pz=3, mass=4)
+        """
+        if sum(x is not None for x in (t, tau)) > 1:
+            raise TypeError("Only one non-None temporal coordinate allowed")
```
jpivarski, line 3099: Same thing.

src/vector/_methods.py
Saransh-cpp: Yes, they do maintain momentum-ness. Right now the generic coordinates are automatically converted to momentum coordinates, but yes, we can explicitly add momentum coordinates as arguments in the future -

```py
In [1]: import vector

In [2]: vector.obj(px=1, py=2).to_Vector4D(z=3, t=4)
Out[2]: MomentumObject4D(px=1, py=2, pz=3, E=4)

In [3]: vector.obj(px=1, py=2).to_Vector4D(eta=3, tau=4)
Out[3]: MomentumObject4D(px=1, py=2, eta=3, mass=4)
```

Saransh-cpp:
Thanks for the review! I will merge this in and tag `v1.0.0`!

Issue #92, opened 2021-04-16
Summing over array dimensions does not work properly


masonproffitt:
Summing doesn't seem supported at all with numpy arrays:

```python
>>> np.sum(vector.array({"x": [1], "y": [1]}))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<__array_function__ internals>", line 5, in sum
TypeError: no implementation found for 'numpy.sum' on types that implement __array_function__: [<class 'vector._backends.numpy_.VectorNumpy2D'>]
>>> vector.array({"x": [1], "y": [1]}).sum()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/user/miniconda3/envs/iris-hep/lib/python3.9/site-packages/numpy/core/_methods.py", line 47, in _sum
    return umr_sum(a, axis, dtype, out, keepdims, initial, where)
TypeError: operand type(s) all returned NotImplemented from __array_ufunc__(<ufunc 'add'>, 'reduce', VectorNumpy2D([(1., 1.)], dtype=[('x', '<f8'), ('y', '<f8')]), axis=None, dtype=None, keepdims=False, where=True): 'VectorNumpy2D'
```

With `ak.Array`, you can get results with `ak.sum()`, but they don't make sense for vectors:

```python
>>> ak.sum(vector.Array({"x": [1], "y": [1]}))
2
```

Using the `axis` parameter works fine for Cartesian coordinates:

```python
>>> ak.sum(vector.Array({"x": [1, 2], "y": [3, 4]}), axis=0)
<Record {x: 3, y: 7} type='{"x": int64, "y": int64}'>
```

but not for any other cases:

```python
>>> ak.sum(vector.Array({"pt": [1, 1], "phi": [0, np.pi]}), axis=0)
<Record {rho: 2, phi: 3.14} type='{"rho": int64, "phi": float64}'>
```

jpivarski:
@lukasheinrich asked about this for Awkward Arrays of Vectors, rather than NumPy arrays of Vectors. (I haven't been able to find the reference.) To support `np.sum`/`ak.sum` for Awkward Arrays of Vectors, we'll need to add a new mechanism for adding behavior to Awkward Array—currently, you can't overload reducers, but that would be a nice feature. To support `np.sum` for NumPy arrays of Vectors, we'd just have to implement `__array_function__`. That's doable.

lukasheinrich:
[...]

PR #347, opened 2023-05-16
feat: implement `sum`, `count`, and `count_nonzero` reductions

agoose77:
@jpivarski @Saransh-cpp this is a WIP initial stab at supporting `sum`, `count`, and `count_nonzero` reducers in vector across both NumPy and Awkward.[^1]

My main query here is around API design; I note that @jpivarski and @henryiii designed the initial prototype, so you'll both likely have stronger feelings on how this deviates from the intended design. In particular, is it reasonable to fit reducers (in this case, only `sum`) into the compute model that is already established? From our type hints here, we don't explicitly encode any expectation that the vectors in compute functions are array like (or indeed satisfy _any_ type checks), so I think this is fine.

Awkward support depends upon scikit-hep/awkward#2458.

[^1]: `count` is Awkward specific, however.


src/vector/backends/awkward.py
```diff
@@ -1978,16 +1978,25 @@ def impl(record: typing.Any) -> typing.Any:
...
+behavior["__numba_typer__", "Momentum2D"] = _numba_typer_Momentum2D
+behavior["__numba_typer__", "Momentum3D"] = _numba_typer_Momentum3D
+behavior["__numba_typer__", "Momentum4D"] = _numba_typer_Momentum4D
+
+behavior["__numba_lower__", "Vector2D"] = _numba_lower
+behavior["__numba_lower__", "Vector3D"] = _numba_lower
+behavior["__numba_lower__", "Vector4D"] = _numba_lower
+behavior["__numba_lower__", "Momentum2D"] = _numba_lower
+behavior["__numba_lower__", "Momentum3D"] = _numba_lower
+behavior["__numba_lower__", "Momentum4D"] = _numba_lower
```
agoose77, line 1993: Was this using the global behavior for a reason?

src/vector/_methods.py
```diff
@@ -3285,6 +3291,14 @@ def isclose(
             )
         return isclose.dispatch(rtol, atol, equal_nan, self, other)
 
+    def sum(self) -> VectorProtocol:
+        from vector._compute.planar import sum
+
+        return sum.dispatch(self)
+
+    def count_nonzero(self) -> ScalarCollection:
+        return self.lib.count_nonzero(self.rho2, axis=-1)
```
agoose77, line 3300: Should this be in a `_compute` module for completeness, or is it acceptable given that we don't need to compute anything new?

src/vector/_methods.py
```diff
@@ -3778,6 +3800,14 @@ def isclose(
             )
         return isclose.dispatch(rtol, atol, equal_nan, self, other)
 
+    def sum(self) -> VectorProtocol:
+        from vector._compute.lorentz import sum
+
+        return sum.dispatch(self)
+
+    def count_nonzero(self) -> ScalarCollection:
+        return self.lib.count_nonzero(self.tau2, axis=-1)
```
agoose77, line 3809: This is a particular interpretation of `nonzero`. It would effectively count non-photons. Some insight into usage may be appreciated here!

src/vector/backends/awkward.py
jpivarski: There's a one `vector.backends.awkward.behavior` dict, but this doesn't get installed in the global `ak.behavior` unless the user opts in (calls `vector.register_awkward()`). Vector's `behavior` dict can be passed around between arrays, so that it applies to some arrays and not others. Just filling up Vector's `behavior` dict is not something I'd call "global."

That said, I'm more inclined toward recommending globally installing behaviors with `vector.register_awkward()` than I used to be. It's global state, but it's assigned once at the beginning of a typical script/library and not changed thereafter. Class and function definitions are also "global state" in this sense, and behaviors are more like class/function definitions than application data. (If it were a compiled language, this would all be precompiled code.) Having this globally installed is better than getting confused about which arrays have Vector behaviors and which don't.

agoose77: I'm on mobile right now, but originally this code was directly using ak.behaviour, which is what i was referring to. It sounds like it wasn't supposed to, based on your reply!

jpivarski: I just noticed the red: you're right! That was a giant mistake. It's not one that anyone would have noticed, but still. These functions are supposed to go into Vector's `behavior` dict, not Awkward's.

agoose77:
Here's a rewrite that implements the reductions for each array backend. I need to cleanup the type hints, but it's mostly there.

src/vector/backends/awkward.py
```diff
@@ -1978,16 +1978,85 @@ def impl(record: typing.Any) -> typing.Any:
...
+    | MomentumArray4D,
+    mask_identity: bool,
+) -> ScalarCollection:
+    mag_2 = array.rho2
+    if isinstance(array, Spatial):
+        mag_2 = mag_2 + array.z**2
+    if isinstance(array, Lorentz):
+        mag_2 = mag_2 + array.t2
+
+    return ak.count_nonzero(mag_2, axis=1)
```
jpivarski, line 2052: Instead of checking that the sum is equal to zero, this could check to see if all components are individually equal to zero (three `== 0` and two `&` on the `rho2`, `z`, and `t2` components). That _feels_ cleaner, though I don't see any way that floating point errors in `+` could cause accidental cancellations in rare cases: all three of these expressions are guaranteed to be positive. It even gives the same result if any are `NaN`. (`NaN` counts as nonzero and `NaN` is infectious: it will spread through the addition.)

So this is okay.

jpivarski (approved):
Just some missing tests: I don't see any

* non-Cartesian examples (though, by looking at the code, I'd expect them to work)
* 2D or 4D examples
* tests of the `np.ndarray.sum` method (does it defer to `np.sum`?)

To answer your question in the PR intro, I think it's very good to not use the `_compute`/dispatch machinery for this, since that machinery was designed for ufunc-like functions, and these are reducer functions. The topology doesn't fit, and twisting the existing machinery to make it fit would complicate it even more than it is. (It's more complicated than I would like, but necessary to achieve the list of goals we wrote at the beginning.)

There are only a few reducers that could be applied to spatial/momentum vectors, `np.sum`, `np.count_nonzero`, and `ak.count` (no NumPy equivalent). Whereas the list of potential ufunc-like functions is endless—I can think of a few more that aren't here—it's hard to imagine any more reducers. Therefore, we don't want a generalized machinery that could obfuscate the implementation of these three, and only three, functions.

So :+1:!

src/vector/backends/awkward.py
agoose77: I initially thought about implementing it that way. Off the top of my head, using the sum is slightly fewer vectorised operations (`array == 0` and 3 `__add__` vs a single `pow(x, 2)` and 3 `__add__`, for Lorentz). It was late when I wrote this, though, because floating-point addition is slower than integer addition and that's enough to make the aesthetics matter too.

On balance, I prefer your suggestion.

Saransh-cpp:
This should also close #92 and maybe https://github.com/iris-hep/func_adl_uproot/issues/98 🎉 

PS: I had a very dirty somewhat working implementation for `np.sum` here - https://github.com/scikit-hep/vector/issues/92#issuecomment-1133962519

agoose77:
> This should also close #92 and maybe [iris-hep/func_adl_uproot#98](https://github.com/iris-hep/func_adl_uproot/issues/98) :tada:
>
> PS: I had a very dirty somewhat working implementation for `np.sum` here - [#92 (comment)](https://github.com/scikit-hep/vector/issues/92#issuecomment-1133962519)

Oops, I missed this! Well, good to know it's needed!

jpivarski:
> everything is moved into Cartesian, and remains in Catesian

Some (many?) of the ufunc-like functions do, too. This is not unique to reducers.

agoose77:
This is ready to go!

jpivarski (approved):
This is ready to go now because [Awkward 2.2.3](https://pypi.org/project/awkward/2.2.3/) is out. (The tests would have failed if they couldn't reach it.)

I did another scan through the diffs, and I'm reminded that this includes some bug-fixes as well as the new feature, but I guess that's alright. (There's assignment of `__numba_typer__` and `__numba_lower__` into Vector's `behavior` instead of `ak.behavior`, and also type hint corrections on `allclose`—booleans, not floats—that I hadn't noticed before.)

I'll merge it.

PR #348, opened 2023-05-22
chore: replace custom definition of np.isclose with numba's np.isclose

### LINKED ISSUE #185: Take Numba's `np.isclose` lowering instead of defining our own.
When numba/numba#7067 gets into a production release of Numba, we can remove this:

https://github.com/scikit-hep/vector/blob/e493e6f77d589571f85dcdf4dc4b61b8ca16f991/src/vector/_backends/numba_object.py#L95-L125

and then make that production release of Numba our minimally supported version. (That version of Numba has to be easily accessible: in pip and conda-forge, not just the "numba" conda channel.)

Then the minor version of Vector should also be bumped, up to 0.9.0.

Saransh-cpp:
## Description

Closes #185 

## Checklist

## Before Merging

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,7 +50,7 @@ awkward = [
 ]
 dev = [
   "awkward>=1.2",
-  'numba>=0.50; python_version < "3.11"',
+  "numba>=0.57",
   "papermill>=2.4",
   "pytest>=6",
   "pytest-cov>=3",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,7 +50,7 @@ awkward = [
 ]
 dev = [
   "awkward>=1.2",
-  "numba>=0.57",
+  'numba>=0.57; python_version > "3.7"',
   "papermill>=2.4",
   "pytest>=6",
   "pytest-cov>=3",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,7 +50,7 @@ awkward = [
 ]
 dev = [
   "awkward>=1.2",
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -50,7 +50,7 @@ awkward = [
 ]
 dev = [
   "awkward>=1.2",
-  'numba>=0.50; python_version < "3.11"',
+  'numba>=0.57; python_version > "3.7"',
```
henryiii, line 53: Numba 0.57 isn’t available for Python 3.7.

Saransh-cpp: Oh wait, saying `python_version > "3.7"`  will install it on Python 3.7.x where x > 0?

henryiii: Ah, no, `python_version` is two digits, so this will actually work. But please use:

```suggestion
  'numba>=0.57; python_version >= "3.8"',
```

As it’s clearer, imo. :)

henryiii:
@jpivarski, should we drop 3.7?

If we wanted to support both for a bit, we could make the workaround conditional instead of immediately dropping it. But I’d only do that with a strong reason.

pyproject.toml
Saransh-cpp: Ah, thanks for the quick explanation!

jpivarski (approved):
Fantastic! Yes, please get rid of our custom `isclose` and require the latest Numba.

Saransh-cpp:
Great, thanks for the quick reviews!

PR #354, opened 2023-06-17
chore: use trusted publisher deployment

henryiii:
This moves publishing over to PyPI's trusted publishing. I'm adding this repo to PyPI now. (Edit: done)


Saransh-cpp (approved):
Thanks, @henryiii!

PR #355, opened 2023-06-19
chore: remove Python 3.7 support

Saransh-cpp:
Following https://github.com/scikit-hep/vector/pull/348#issuecomment-1557596467, this PR removes the support of Python 3.7 from vector. I'm still not completely sure if there was a concensus regarding this but I'll be happy to close this PR if we do want to support Python 3.7.

## Description

## Checklist

## Before Merging

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@
 [![LICENSE][license-badge]][license-link]
 [![Scikit-HEP][sk-badge]][sk-link]
 
-Vector is a Python 3.7+ library for 2D, 3D, and [Lorentz vectors](https://en.wikipedia.org/wiki/Special_relativity#Physics_in_spacetime), especially _arrays of vectors_, to solve common physics problems in a NumPy-like way.
+Vector is a Python 3.8+ library (Python 3.6 and 3.7 supported till `v0.9.0` and `v1.0.0`, respectively) library for 2D, 3D, and [Lorentz vectors](https://en.wikipedia.org/wiki/Special_relativity#Physics_in_spacetime), especially _arrays of vectors_, to solve common physics problems in a NumPy-like way.
 
 Main features of Vector:
 
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -28,6 +28,11 @@ jobs:
     needs: [dist]
     runs-on: ubuntu-latest
     if: github.event_name == 'release' && github.event.action == 'published'
+    environment:
+      name: pypi
+      url: https://pypi.org/p/vector
+    permissions:
+      id-token: write
 
     steps:
       - uses: actions/download-artifact@v3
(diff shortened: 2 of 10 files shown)


README.md
```diff
@@ -17,7 +17,7 @@
 [![LICENSE][license-badge]][license-link]
 [![Scikit-HEP][sk-badge]][sk-link]
 
-Vector is a Python 3.7+ library for 2D, 3D, and [Lorentz vectors](https://en.wikipedia.org/wiki/Special_relativity#Physics_in_spacetime), especially _arrays of vectors_, to solve common physics problems in a NumPy-like way.
+Vector is a Python 3.8+ library (Python 3.6 and 3.7 supported till `v0.9.0` and `v1.0.0`, respectively) library for 2D, 3D, and [Lorentz vectors](https://en.wikipedia.org/wiki/Special_relativity#Physics_in_spacetime), especially _arrays of vectors_, to solve common physics problems in a NumPy-like way.
```
Saransh-cpp, line 20:
```suggestion
Vector is a Python 3.8+ (Python 3.6 and 3.7 supported till `v0.9.0` and `v1.0.0`, respectively) library for 2D, 3D, and [Lorentz vectors](https://en.wikipedia.org/wiki/Special_relativity#Physics_in_spacetime), especially _arrays of vectors_, to solve common physics problems in a NumPy-like way.
```

Saransh-cpp, line 20:
```suggestion
Vector is a Python 3.8+ library (Python 3.6 and 3.7 supported till `v0.9.0` and `v1.0.0`, respectively) for 2D, 3D, and [Lorentz vectors](https://en.wikipedia.org/wiki/Special_relativity#Physics_in_spacetime), especially _arrays of vectors_, to solve common physics problems in a NumPy-like way.
```

henryiii:
Ruff's config still has:
```toml
target-version = "py37"
```

Saransh-cpp:
Oops, fixed it, thank you!

henryiii:
@jpivarski thoughts? Should this go in 1.1? I’m thinking it might be good to do now since 1.1 doesn’t support numba on 3.7, while 1.0 does.

jpivarski:
Yes, it could be good to do all of the dependency version threshold changes in 1.1, instead of using multiple minor versions or using up numbers.

It's a tighter set of version constraints for someone who wants #347, but it doesn't sound like a likely scenario. ("I want `ak.sum` to work on Vectors, but I'm using Python 3.7.")

I'd say go ahead and add it to Vector 1.1.

henryiii:
It's not just about using up numbers, this release is actually a downgrade for Python 3.7 users - they now can't use numba with vector anymore if we don't bump the Python version now. Using `ak.sum` is a new feature and won't break existing code like removing numba could.

henryiii (approved):

PR #371, opened 2023-08-25
chore: remove license string (not standard)

henryiii:
This is not following PEP 621, and we already have the canonical license classifier.


Saransh-cpp (commented):
Thanks! Went unnoticed.

PR #388, opened 2023-10-03
support Python 3.12

Saransh-cpp:
## Description

## Checklist

## Before Merging

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,6 +39,7 @@ jobs:
           - "3.9"
           - "3.10"
           - "3.11"
+          - "3.12"
     name: Python ${{ matrix.python-version }} - Light
     steps:
       - uses: actions/checkout@v4
@@ -67,6 +68,7 @@ jobs:
           - "3.9"
           - "3.10"
           - "3.11"
+          - "3.12"
     name: Python ${{ matrix.python-version }} - Awkward v1
     steps:
       - uses: actions/checkout@v4
@@ -98,6 +100,7 @@ jobs:
           - "3.9"
           - "3.10"
           - "3.11"
+          - "3.12"
     name: Python ${{ matrix.python-version }} - Awkward v2
     steps:
       - uses: actions/checkout@v4
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## Unreleased
+
+#### Maintenance
+
+- chore: support Python 3.12 [#388][]
+
+[#388]: https://github.com/scikit-hep/vector/pull/388
+
 ## Version 1.1
 
 ### Version 1.1.1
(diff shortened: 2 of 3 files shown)


henryiii:
We need to make numba conditional on Python version. It’s always slow.

Saransh-cpp:
Thanks, @henryiii!

The awkward (v1) - numba backend has been failing with the numba `v0.58` release. Looking at this now, should I try debugging this, or should we entirely drop awkward v1 support from vector? I believe everyone must have migrated to awkward v2 by now, and it should be okay to stop supporting awkward v1.

cc: @jpivarski

jpivarski:
It's not the case that everyone has migrated to Awkward 2 because Coffea 2023 (which depends on Awkward 2 instead of Awkward 1) is still in release-candidate.

However, when Coffea switches from its built-in vectors to Vector, it will be in the 2023 series, so I think there wouldn't be a version conflict to drop Awkward 1 from Vector.

But this is a bigger problem than Vector anyway: Awkward 1 should either fix its Numba 0.58 support or cap it (which would be difficult because it's a soft/runtime dependency).

jpivarski:
Also, looking at the error message,

```
FAILED tests/backends/test_awkward_numba.py::test - numba.core.errors.LoweringError: Failed in nopython mode pipeline (step: native lowering)
Failed in nopython mode pipeline (step: nopython frontend)
Code using Numba extension API maybe depending on 'old_style' error-capturing, which is deprecated and will be replaced by 'new_style' in a future release. See details at https://numba.readthedocs.io/en/latest/reference/deprecation.html#deprecation-of-old-style-numba-captured-errors
Exception origin:
  File "/opt/hostedtoolcache/Python/3.11.5/x64/lib/python3.11/site-packages/awkward/_connect/_numba/arrayview.py", line 877, in generic
    raise TypeError(


File "tests/backends/test_awkward_numba.py", line 23:
    def extract(x):
        return x[2][0]
        ^

During: lowering "$20binary_subscr.4 = static_getitem(value=$8binary_subscr.2, index=0, index_var=$const18.3, fn=<built-in function getitem>)" at /home/runner/work/vector/vector/tests/backends/test_awkward_numba.py (23)
```

I think this might be a warning, not an error, so Awkward 1 doesn't get into a true conflict with Numba until some version after 0.53. Something should be done now, anyway.

Saransh-cpp:
Thanks for the explanation, @jpivarski!

jpivarski:
A new Awkward 1 version is ready: https://github.com/scikit-hep/awkward/discussions/2737

Saransh-cpp:
Everything works, thanks, @jpivarski!

Saransh-cpp:
I should have asked this before merging, but should we bump Awkward's minimum version to `v1.10.5` and create a new `v1.2` release?

https://github.com/scikit-hep/vector/blob/30e19f967099ed59d190f5d24e34d2abb8171a9b/pyproject.toml#L45-L57

henryiii:
There were no code changes, right? It's just adding CI and the trove classifier?

henryiii:
I don't think we need to bump the minimum Awkward, people get the highest version unless something caps, in which case bumping the minimum makes the solve impossible. I'd only bump the minimum if vector itself doesn't support 1.2 of if we have a lower bound dependency on something 1.2 doesn't work with (Numba's not a hard dep), not if 1.2 doesn't work anymore with the latest dependencies.

henryiii:
I don't see any code changes here: https://github.com/scikit-hep/vector/compare/v1.1.1...main

So at best, if you want the updated classifier, we could so a 1.1.2, or maybe just a 1.1.1.post1.

Saransh-cpp:
That makes sense, I will create a 1.1.1.post1 (too many ones) release. Thanks!

Issue #422, opened 2024-02-16
Revisiting vector dimension projecting/embedding


jpivarski:
We had another conversation today that reopens issue #412 (implemented by PR #413). When adding/subtracting vectors of different spatial/momentum dimensions—i.e. 2D vs 3D vs 4D, not array shape—there are three possibilities:

1. the low-D vector gets promoted to the high-D dimensionality, filling the missing slots with zeros (so that they don't contribute to the sum)
2. the high-D vector gets demoted to the low-D dimensionality, ignoring the information in its extra dimensions
3. it's a TypeError that explains how to promote or demote in the error message.

@Saransh-cpp, @nsmith-, @iasonkrom, @lgray, and @martindurant were involved in that discussion. Nick restated his opinion in #412 (option 2), but it seemed that the majority were in favor of option 3, as long as it's easy to promote/demote.

Since spatial/momentum dimensionality is expressed as record fields, high-D vectors are subtypes of low-D vectors (high-D vectors have a superset of the fields that a low-D vector has), and high-D → low-D is _upcasting_ (removing those extra fields from the high-D vector). Upcasting requires no extra information, but downcasting requires extra information: the values of the new fields that are being added (though we'd make them 0, in a way that is neutral for addition and subtraction).

There's no ambiguity about _which_ dimensions are dropped or added when going between low-D and high-D, since our vectors have a special 3rd component (cylindrically symmetric, not spherically symmetric) and a special 4th component: 2D is always azimuthal, the 3rd dimension in 3D is always longitudinal, and the 4th dimension in 4D is always temporal.

If we're going with option 3, the methods that _project_ high-D → low-D and the methods that _embed_ low-D → high-D have to be short and not clutter up a mathematical expression (like NumPy's `.T`, just two characters to be just-visible-enough). It would be nice to have a single word for this, instead of "project" and "embed," although those are the correct mathematical terms.

We could have `.to_2D()`, `.to_3D()`, `.to_4D()`, since we don't have an ambiguity about which dimensions are being added or removed. These would be shorter synonyms of functions that already exist to do this.

Iason suggested `.like(other_array)`, so that we don't need to specify which "D". I like that and would +1 that. (If you want to specify which "D", there are already functions for that; the new convenience is being able to fit it to a given `other_array`.) The error message in option 3 could be

```
TypeError: cannot add or subtract vectors of different dimensionality; use

    a.like(b) + b

or

    a + b.like(a)

to project or embed one of the vectors to match the other's dimensionality
[...]

PR #423, opened 2024-02-19
feat: allow momentum coords in to_Vector*D methods + cleanup

### LINKED ISSUE #382: Allow momentum type coordinates in `to_Vector*D` methods
If so, then a user might justifiably expect to pass `pz` instead or `z`, but only in the momentum case. Just something to consider, at this point.

When this gets extended to `to_Vector4D`, the time component has a lot of momentum synonyms. The geometric names for the two possible temporal coordinates are `t` and `tau`, but the momentum names are `energy` and `mass`, `E` and `M`, and (I think) `e` and `m`. There was a discussion once about including the lowercase as well as the uppercase.

_Originally posted by @jpivarski in https://github.com/scikit-hep/vector/pull/319#discussion_r1115886750_

Saransh-cpp:
## Description

Fixes #382

Paves the way for #422 (make the conversions more user-friendly)

cc: @nsmith-, @iasonkrom, @lgray

## Checklist

## Before Merging


tests/compute/test_demotion.py
```diff
@@ -0,0 +1,164 @@
+# Copyright (c) 2019-2023, Jonas Eschle, Jim Pivarski, Eduardo Rodrigues, and Henry Schreiner.
```
Saransh-cpp, line 1:
```suggestion
# Copyright (c) 2019-2024, Jonas Eschle, Jim Pivarski, Eduardo Rodrigues, and Henry Schreiner.
```

src/vector/_methods.py
```diff
@@ -3209,49 +3214,55 @@ def to_Vector4D(
...
         theta: float | None = None,
         eta: float | None = None,
         t: float | None = None,
+        e: float | None = None,
+        E: float | None = None,
+        energy: float | None = None,
         tau: float | None = None,
+        m: float | None = None,
+        M: float | None = None,
+        mass: float | None = None,
```
jpivarski, line 3227: If somebody passes both `z` and `pz`, or both `z` and `theta` for that matter, does the function complain? It shouldn't just pick one and use it—providing (possibly) conflicting arguments should be an error, to prevent silent mistakes.

jpivarski (approved):
I remember the discussion on this and I must have missed the review request.

It's a good extension, though I'm worried about the possibility that a user would pass both `z` and `pz` to one of these functions (see below).

src/vector/_methods.py
Saransh-cpp: Thanks for the review! Yes, the functions was checking for multiple generic coordinates, and I've added momentum coordinated to that check too.

src/vector/_methods.py
```diff
@@ -3209,49 +3214,55 @@ def to_Vector4D(
...
+            t_value = next(coord for coord in (tau, m, M, mass) if coord is not None)
+        elif any(coord is not None for coord in (t, e, E, energy)):
+            t_value = next(coord for coord in (t, e, E, energy) if coord is not None)
 
         l_value = 0.0
         l_type: type[Longitudinal] = LongitudinalZ
-        if z is not None:
-            l_value = z
+        if any(coord is not None for coord in [z, pz]):
+            l_value = next(coord for coord in [z, pz] if coord is not None)
```
Saransh-cpp, line 3265:
```suggestion
        if any(coord is not None for coord in (z, pz)):
            l_value = next(coord for coord in (z, pz) if coord is not None)
```

src/vector/_methods.py
```diff
@@ -3163,34 +3168,34 @@ def to_Vector3D(
...
-                "At most one longitudinal coordinate (`z`, `theta`, or `eta`) may be assigned (non-None)"
+                "At most one longitudinal coordinate (`z`/`pz`, `theta`, or `eta`) may be assigned (non-None)"
             )
 
         l_value = 0.0
         l_type: type[Longitudinal] = LongitudinalZ
-        if z is not None:
-            l_value = z
+        if any(coord is not None for coord in [z, pz]):
+            l_value = next(coord for coord in [z, pz] if coord is not None)
```
Saransh-cpp, line 3198:
```suggestion
        if any(coord is not None for coord in (z, pz)):
            l_value = next(coord for coord in (z, pz) if coord is not None)
```

PR #426, opened 2024-02-21
(WIP) feat: error out on operations on vectors of different geometric dimensions + new `like` method

Saransh-cpp:
## Description

Will rebase once #423 is merged.

## Checklist

## Before Merging


Saransh-cpp:
Hi @jpivarski, @nsmith-, @iasonkrom, @lgray

I've noticed that some operations require the vectors to be of different dimensions; for instance, `boost_beta3` (the failing test).

Is there a specific set of operations that should error out when performed on vectors of different dimensionality, or maybe I should ask about a particular set of operations that should _not_ error?

The `like` method and the `TypeError` are working well in this PR! (see tests)

iasonkrom:
I would assume that those that require different dimensionality are fewer so.... the second option? Y'all may have better opinions though.

Saransh-cpp:
> Some methods do require arguments of a particular type to be meaningful, such as `LorentzVector.boost_beta3` should only accept a 3-component monentum vector.

Ah, this can be easily highlighted by updating the type signature, so I think an error in such situations will be redundant?
```diff
--- a/src/vector/_methods.py
+++ b/src/vector/_methods.py
@@ -3818,7 +3818,7 @@ class Lorentz(Spatial, VectorProtocolLorentz):
         return boost_p4.dispatch(self, p4)
 
     def boost_beta3(
-        self: SameVectorType, beta3: VectorProtocolSpatial
+        self: SameVectorType, beta3: MomentumProtocolSpatial
     ) -> SameVectorType:
         from vector._compute.lorentz import boost_beta3
```

Saransh-cpp:
Thinking about this again, do we want strict checks in each method/operation or do we want strict checks only in the methods/operations that can accept every type of vector (`v: VectorProtocol`).

For instance, `add` has the following signature:
```py
def add(self, other: VectorProtocol) -> VectorProtocol:
```

Therefore, I have added a type check in this method. But, `cross` has the following updated signature in this PR:
```py
def cross(self, other: VectorProtocolSpatial) -> VectorProtocolSpatial:
```

and the following works:
```py
In [1]: import vector

In [2]: vector.obj(x=0.1, y=0.2, z=0.3, t=10).cross(vector.obj(x=0.4, y=0.5, z=0.6, t=20))
Out[2]: VectorObject3D(x=-0.03, y=0.06, z=-0.030000000000000013)
```

From an API design point of view, should we check the type of an argument even after providing a type hint? It feels redundant to me. Maybe the implementation of `cross` should be changed such that it does not generalize to 4D vectors, instead of explicitly checking the type of the vectors before executing the logic.

cc: @jpivarski

README.md
```diff
@@ -531,11 +531,17 @@ vector.obj(pt=1, phi=1.3, eta=2, mass=5).deltaR(
...
-vector.obj(x=1, y=2, z=3) - vector.obj(x=1, y=2)
-vector.obj(x=1, y=2, z=0).is_parallel(vector.obj(x=1, y=2))
+v1 = vector.obj(x=1, y=2, z=3)
+v2 = vector.obj(x=1, y=2)
+
+v1 - v2.like(v1), v1.like(
+    v2
+) - v2, v1 - v2.to_xyz(), v1.to_xy() - v2, v1 - v2.to_Vector3D(
+    z=3
+), v1.to_Vector2D() - v2
```
jpivarski, line 544: This might be required for black-compliant formatting, but it makes what would be easy to read into something difficult to read.

Do you need the commas?

```suggestion
v1 - v2.like(v1)
v1.like(v2) - v2
v1 - v2.to_xyz()
v1.to_xy() - v2
v1 - v2.to_Vector3D(z=3)
v1.to_Vector2D() - v2
```

Should there be some explanatory text between these?

jpivarski (approved):
The new `like` method is documented, but there should also be a short comment about it in the error messages that result from trying to add/subtract/compare vectors of different dimensionality, since the `like` method is how users would fix those errors.

The exclusion of implicit projection/embedding and inclusion of a new `like` method together form a single project of asking users for clarity about whether they want to project or embed. Some text like

```
x + y   →   x.like(y) + y   or   x + y.like(x)
```

would indicate where the user is being asked for their opinion.

src/vector/_methods.py
```diff
@@ -4143,6 +4180,26 @@ def dim(v: VectorProtocol) -> int:
...
+                a.like(b).{operation}(b)
+
+            or
+
+                a.{operation}(b.like(a))
+
+            or the binary operation equivalent to project or embed one of the vectors
+            to match the other's dimensionality
+            """
+        )
```
Saransh-cpp, line 4200: Thanks for the review, @jpivarski! The error message here displays the usage of the new `like` method. Should this be changed in any way?

jpivarski (approved):
This is great! I approve!

Having a utility function for that error message is a good idea; it seems to be needed in a lot of methods.

Saransh-cpp:
Great, thanks for the review!

PR #429, opened 2024-02-28
feat: add support for dask-awkward arrays in vector constructors

### LINKED ISSUE #301: Support for `dask-awkward` arrays
### Describe the potential feature

Support for vectors and vector operations using arrays from [`dask-awkward`](https://github.com/dask-contrib/dask-awkward/). I assume this is the correct repository for this issue, since by parallel construction there would need to be something like `vector.dask`.

### Motivation

This is of course important for many HEP uses of `uproot.dask`. Among other places, this is needed for adoption of Uproot v5 and Awkward Array v2 by `func-adl-uproot` and therefore the Uproot ServiceX transformer.

### Possible Implementation

_No response_

Saransh-cpp:
## Description

Fixes #301 

## Checklist

## Before Merging

### DIFF
--- a/src/vector/backends/awkward_constructors.py
+++ b/src/vector/backends/awkward_constructors.py
@@ -315,6 +315,7 @@ def Array(*args: typing.Any, **kwargs: typing.Any) -> typing.Any:
 
     if vector._is_awkward_v2:
         import dask_awkward  # type: ignore[import-not-found]
+
         ak_array_types += (dask_awkward.Array,)
 
     if not isinstance(args[0], ak_array_types):
--- a/src/vector/backends/awkward_constructors.py
+++ b/src/vector/backends/awkward_constructors.py
@@ -328,7 +328,7 @@ def Array(*args: typing.Any, **kwargs: typing.Any) -> typing.Any:
     is_momentum, dimension, names, arrays = _check_names(akarray, fields)
 
     # don't execute for dask_awkward arrays
-    if isinstance(args[0], (awkward.Array, list)):
+    if isinstance(args[0], (awkward.Array, list, numpy.ndarray)):
         needs_behavior = not vector._awkward_registered
         for x in arrays:
             if needs_behavior:
--- a/src/vector/backends/awkward_constructors.py
+++ b/src/vector/backends/awkward_constructors.py
@@ -311,7 +311,7 @@ def Array(*args: typing.Any, **kwargs: typing.Any) -> typing.Any:
     import vector
     import vector.backends.awkward
 
-    ak_array_types = (awkward.Array,)
+    ak_array_types: tuple[typing.Any, ...] = (awkward.Array,)
 
     if vector._is_awkward_v2:
         import dask_awkward  # type: ignore[import-not-found]
--- a/src/vector/backends/awkward_constructors.py
+++ b/src/vector/backends/awkward_constructors.py
@@ -307,11 +307,15 @@ def Array(*args: typing.Any, **kwargs: typing.Any) -> typing.Any:
     integers or floating-point numbers.
     """
     import awkward
+    import dask_awkward  # type: ignore[import-not-found]
 
     import vector
     import vector.backends.awkward
 
-    akarray = awkward.Array(*args, **kwargs)
+    if not isinstance(args[0], (awkward.Array, dask_awkward.Array)):
+        akarray = awkward.Array(*args, **kwargs)
+    else:
+        akarray = args[0]
     array_type = akarray.type
 
     if not _is_type_safe(array_type):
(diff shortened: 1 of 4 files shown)


src/vector/backends/awkward_constructors.py
```diff
@@ -320,24 +330,15 @@ def Array(*args: typing.Any, **kwargs: typing.Any) -> typing.Any:
...
-        if needs_behavior:
-            if x.behavior is None:
-                x.behavior = vector.backends.awkward.behavior
-            else:
-                x.behavior = dict(x.behavior)
-                x.behavior.update(vector.backends.awkward.behavior)
-        else:
-            x.behavior = None
-        needs_behavior = False
-
```
Saransh-cpp, line 334: I am not sure if this piece of code is still required. If it is, is there a way to set `behavior` for `dask_awkward` arrays?

jpivarski: This code block looks important. We don't want to add a giant `behavior` dict to every array if the relevant behaviors have already been installed globally.

All of the `ak.*` functions can be used on dask-awkward arrays[^1], so the code block that you have below, which builds the vector by calling `ak.with_name` passing the `behavior` gets it into the new dask-awkward array.

[^1]: Any `ak.*` function that takes `ak.Arrays` as arguments and returns an `ak.Array` can also be a function that takes `dak.Arrays` as arguments and returns a `dak.Array`. Awkward now has an overload mechanism like NumPy's.

jpivarski (approved):
This is good, but I think you need the block that checks to see if the Vector behaviors have been installed globally before deciding whether to attach them all to the output array.

src/vector/backends/awkward_constructors.py
Saransh-cpp: Thank you! Given that `behavior` of a `dask_awkward.Array` has no setter, should I use conditional statements here to preserve the code block (subsequently using `if` while passing `behavior` in `awkward.with_name`) or is there a better way to do this?
```py
if vector._is_awkward_v2 and not isinstance(args[0], dask_awkward.Array):
    needs_behavior = not vector._awkward_registered
    for x in arrays:
        if needs_behavior:
            if x.behavior is None:
                x.behavior = vector.backends.awkward.behavior
            else:
                x.behavior = dict(x.behavior)
                x.behavior.update(vector.backends.awkward.behavior)
        else:
            x.behavior = None
        needs_behavior = False

assert 2 <= dimension <= 4, f"Dimension must be between 2-4, not {dimension}"

return awkward.with_name(
    awkward.zip(
        dict(__builtins__["zip"](names, arrays)),  # type: ignore[index]
        depth_limit=akarray.layout.purelist_depth,
    ),
    _recname(is_momentum, dimension),
    behavior=vector.backends.awkward.behavior
    if vector._is_awkward_v2 and isinstance(args[0], dask_awkward.Array))
    else None,
)
```

Saransh-cpp:
@jpivarski, could you please take a quick look at this again? I'll merge this today. Thanks!

src/vector/backends/awkward_constructors.py
```diff
@@ -311,7 +311,17 @@ def Array(*args: typing.Any, **kwargs: typing.Any) -> typing.Any:
     import vector
     import vector.backends.awkward
 
-    akarray = awkward.Array(*args, **kwargs)
+    ak_array_types: tuple[typing.Any, ...] = (awkward.Array,)
+
+    if vector._is_awkward_v2:
+        import dask_awkward  # type: ignore[import-not-found]
+
+        ak_array_types += (dask_awkward.Array,)
```
jpivarski, line 319: This is an example of why I think we should drop Awkward 1 support soon. Generally, the typing code should be declarative.

By the way, I think you can declare a `dask_awkward.Array` dependency without importing `dask_awkward` by using a string: `"dask_awkward.Array"`. I'm not 100% sure about typing syntax; maybe this only works past a particular Python version. However, we _want_ to be able to allow some types without necessarily importing them (a start-up time cost) or even having them installed (a dependency-complexity cost).

jpivarski (approved):
The implementation is right: you just use `ak.with_name` and it does the right thing for `dask_awkward.Array`s.

I'm a little unsure of the typing, and this goal of having mypy be able to type-check the library and user code might be competing with other goals of simplicity and not getting tangled in dependencies that may be unnecessary at runtime.

Both goals should be attainable. I'd like someone with more Python typing experience to check it, maybe @henryiii? (If you're not too busy?)

src/vector/backends/awkward_constructors.py
Saransh-cpp: I see, yes, that makes sense. I could remove awkward v1 support in this PR itself, but it might be better to do that in the `v1.3.1`/`v1.4.0` release. [PEP 484 mentions forward referencing](https://peps.python.org/pep-0484/#forward-references) to deal with such cases, but that will only be helpful while adding type hints.

I have pushed a commit rewriting the logic for handling `dask_awkward` arrays. The implementation does not require importing `dask_awkward`. Please let me know if that works!

jpivarski: Dropping a major dependency version should happen at some significant version threshold in Vector. Maybe 1.3.x → 1.4.0, or maybe 1.x → 2.0.0.

I like the idea of 2.0.0 (Uproot increases major version when Awkward increases major version), but this has scared users in the past: seeing Uproot 4 → 5 made some people think that the changes in Uproot were major, rather than the changes in its dependencies.

So maybe 1.4.0, yeah.

Saransh-cpp: Thanks! I'll drop awkward v1 in the next release.

src/vector/backends/awkward_constructors.py
```diff
@@ -320,24 +327,29 @@ def Array(*args: typing.Any, **kwargs: typing.Any) -> typing.Any:
 
     is_momentum, dimension, names, arrays = _check_names(akarray, fields)
 
-    needs_behavior = not vector._awkward_registered
-    for x in arrays:
-        if needs_behavior:
-            if x.behavior is None:
-                x.behavior = vector.backends.awkward.behavior
+    # don't execute for dask_awkward arrays
+    if isinstance(args[0], (awkward.Array, list)):
```
Saransh-cpp, line 331:
```suggestion
    if isinstance(args[0], (awkward.Array, list, numpy.ndarray)):
```

PR #511, opened 2024-10-02
ci: 3.13, attestations, and some modernization

henryiii:
This does some modernization, and adds 3.13 testing and most importantly, the beautiful 3.13 classifier. :)


.github/workflows/ci.yml
```diff
@@ -70,23 +70,27 @@ jobs:
...
         with:
           python-version: ${{ matrix.python-version }}
+          allow-prereleases: true
 
-      - name: Requirements check
-        run: python -m pip list
+      - uses: astral-sh/setup-uv@v3
+
+      - name: Install nox
+        run: uv tool install nox
```
Saransh-cpp, line 86: Should we `uvx`/install `nox` using `uv` consistently across this file, or is there a reason to switch between these 2 in different jobs?

Saransh-cpp (approved):
Thanks for this, @henryiii! :)

.github/workflows/ci.yml
henryiii: I think one just used it on more lines. We can normalize to either one if that's better.

Saransh-cpp: Ah, I see, that makes sense. Leaving it as it is looks okay. Thanks!

henryiii:
@Saransh-cpp , it would be good to get patch releases before 3.13 is released, on Tuesday!

Saransh-cpp:
On it 🫡

PR #558, opened 2025-01-21
fix: lib for typetracers

pfackeldey:
This PR fixes https://github.com/scikit-hep/coffea/pull/1195 (the current vector release 1.6.0 is broken for typetracers) 

The problem was that numpy (`lib`) functions were directly called on `TypeTracerArray`s. Instead, we should've used the corresponding `nplike` implementation if it exists. This is only implemented for the typetracer backend as all other backends have already a dedicated dispatch mechanism.

In order to ensure we're catching all errors with this I modified `tests/backend/test_awkward.py` to run every test also for the typetracer backend.

If this PR looks good and gets merged we should make a new release (and maybe yank the previous one?).

Sorry for the inconvenience!

(cc @lgray)

### DIFF
--- a/tests/backends/test_awkward.py
+++ b/tests/backends/test_awkward.py
@@ -5,6 +5,7 @@
 
 from __future__ import annotations
 
+import functools
 import importlib.metadata
 import numbers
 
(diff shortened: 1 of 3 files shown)


tests/backends/test_awkward.py
```diff
@@ -26,97 +27,120 @@
...
-def test_dimension_conversion():
+def _assert(bool1, bool2, backend):
+    # for 'typetracer' backend we pass through (no-op)
+    # for 'cpu' we want to make sure the bool evaluates to True
+    if backend == "cpu":
+        if isinstance(bool1, ak.Array):
+            bool1 = bool1.tolist()
+        if isinstance(bool2, ak.Array):
+            bool2 = bool2.tolist()
+        else:
```
henryiii, line 38: This just falls through and does nothing if bool2 is a ak.Array?

tests/backends/test_awkward.py
```diff
@@ -26,97 +27,120 @@
...
-        ]
+        ],
+        backend=backend,
     )
+
     # test alias
-    assert ak.all(vec.to_3D(z=1).z == 1)
-    assert ak.all(vec.to_3D(eta=1).eta == 1)
-    assert ak.all(vec.to_3D(theta=1).theta == 1)
+    assert_backend(ak.all(vec.to_3D(z=1).z == 1), True)
```
henryiii, line 56: Wouldn't it be a lot clearer if these were just inside an `if backend == "cpu"` block? This really hides what's happening and that it's a simi-no-op? Though I guess it's still computing them, it's just not checking to see if it's true? Could it be equal to a `backend == "cpu"` constant or something?

tests/backends/test_awkward.py
pfackeldey: it calls `.tolist()` if it's an `ak.Array`

tests/backends/test_awkward.py
pfackeldey: For the typetracer backend I want the expressions still be computed, but we can't compare it to numerical values, which is why the assertion is essentially a no-op. This is to make sure that the computation does not fail for the typetracer backend (without checking its result).

pfackeldey: We might want to compare the `ak.forms.Form` in the typetracer case, but that's kind of already checked generously in awkward-array itself - so not really useful. 
So this is just to make sure a vector computation does not fail for the typetracer backend.

Saransh-cpp:
The release on PyPI can be yanked by @henryiii or @jpivarski. I can create a v1.6.1 release and add a note on the v1.6 release as soon as this PR is merged.

henryiii (approved):

henryiii:
Yanked 1.6.0.

Issue #531, opened 2024-11-26
Replace uncompyle6 test with `inspect` and `ast.parse` or just drop it (Python 3.8 is end-of-lifed)


jpivarski:
Although we can keep checking [uncompyle6 releases](https://github.com/rocky/python-uncompyle6/releases), I think the state of the art is fixed at

> Although we do not decompile bytecode greater than 3.8, code supports running from up to 3.12.

Our [test_compute_features.py](https://github.com/scikit-hep/vector/blob/main/tests/test_compute_features.py) test uses uncompyle6 to get an AST of all the functions in the `compute` submodule and verify that they restrict themselves to a small set of features, so that there are fewer limitations on new backends. The new SymPy backend is also helping to keep the `compute` functions general, since the SymPy backend makes fewer assumptions than anything else (vector components aren't even numerical!), but it would be nice to keep directly checking.

[Python 3.8 support ended over a month ago.](https://endoflife.date/python) But, rather than dropping the AST checks, what if we use Python's `inspect` module to find the source code lines and re-parse them with `ast.parse`?

Aha! It's because many of them are auto-generated. For instance, [this for loop](https://github.com/scikit-hep/vector/blob/608da2d55a74eed25635fd408d1075b568773c99/src/vector/_compute/lorentz/add.py#L170-L187) generates 48 functions via composition. Python's `inspect` can't find the source code because there never was any source code for these functions—they're made on-the-fly.

So if we want to preserve the [test_compute_features.py](https://github.com/scikit-hep/vector/blob/main/tests/test_compute_features.py) tests (past the point when Python 3.8 can no longer be accessed in GitHub Actions), we'd have to generate source code text for these functions. One way to do that would be to use uncompyle6 to get AST and then use `ast.unparse` (or similar functionality from uncompyle6) to make source code text that we'll check into git instead of the generating functions. Or modify the generating functions to produce source code text directly. But, either way, we'd switch to managing 48 (not a big number) functions, rather than the for loop that generates them. As a side-bonus, error messages would report line numbers that exist in source code that you can look up.

Alternatively, we'll just remove [test_compute_features.py](https://github.com/scikit-hep/vector/blob/main/tests/test_compute_features.py) when it becomes impossible to install Python 3.8 (in an easy way, without deadsnakes...)

PR #621, opened 2025-08-08
fix: correctly infer the numpy-like module

pfackeldey:
This PR fixes the handling of `lib` for non-NumPy and non-TypeTracer backends with awkward array.

fyi @MoAly98

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -59,6 +59,7 @@ optional-dependencies.dev = [
   "pytest-cov>=3",
   "pytest-doctestplus",
   "sympy",
+  "jax",
 ]
 optional-dependencies.docs = [
   "awkward>=2",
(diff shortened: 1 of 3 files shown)


pfackeldey:
(btw, I think we should deprecate python 3.8 by now...)

Saransh-cpp:
> (btw, I think we should deprecate python 3.8 by now...)

We have 3.8 in CI mostly because of #531. I haven't had the time to look at it yet, but I guess we can keep running 3.8 in the CI and think about it once it can no longer be accessed in GH Actions.

pfackeldey:
@Saransh-cpp this is independent of python 3.8. In case you're happy with it, this could be merged from my side.

Saransh-cpp:
Thanks, @pfackeldey! Looks good to me.

PR #622, opened 2025-08-08
fix: drop python 3.8; it's EOL

pfackeldey:
Dropping python 3.8, it's EOL.


pfackeldey:
ah, my fault!! hm, ok, then we need to rethink this :/

CONTRIBUTING.md
```diff
@@ -207,7 +207,7 @@ nox -s tests        # test with all extra dependencies on all available Python v
 nox -s tests        # run doctests on all available Python versions
 nox -s coverage     # run the tests session and generate a coverage report on all available Python versions
 nox -s notebooks    # test notebooks on the default Python version
-nox -s disassemble  # check compute functions on Python 3.8
+nox -s disassemble  # check compute functions on Python 3.9
```
henryiii, line 210:
```suggestion
```

PR #637, opened 2025-09-16
feat: add PyTree support for vector objects

nsmith-:
## Description

See https://blog.scientific-python.org/pytrees/ for the rationale for these functions.

As a demonstration, here's a projectile motion with air resistance simulation using vector and scipy, showing in particular the use of `vector.pytree.flatten` to wrap the SciPy ODE solver:

```python
import hepunits as u
import matplotlib.pyplot as plt
import numpy as np
from scipy.integrate import solve_ivp

import vector


def wrapped_solve(fun, t_span, y0, t_eval):
    flat_y0, treedef = vector.pytree.flatten(y0)

    def flat_fun(t, flat_y):
        state = vector.pytree.unflatten(treedef, flat_y)
        dstate_dt = fun(t, state)
        flat_dstate_dt, _ = vector.pytree.flatten(dstate_dt)
        return flat_dstate_dt

    flat_solution = solve_ivp(
        fun=flat_fun,
        t_span=t_span,
        y0=flat_y0,
        t_eval=t_eval,
    )
    return vector.pytree.unflatten(treedef, flat_solution.y)


gravitational_acceleration = vector.obj(x=0.0, y=-9.81) * (u.m / u.s**2)
air_density = 1.25 * u.kg / u.meter3
drag_coefficient = 0.47  # for a sphere
object_radius = 10 * u.cm
object_area = np.pi * object_radius**2
object_mass = 1.0 * u.kg


def force(position: vector.Vector2D, momentum: vector.Vector2D):
    gravitational_force = object_mass * gravitational_acceleration
    speed = momentum.rho / object_mass
    drag_magnitude = 0.5 * air_density * speed**2 * drag_coefficient * object_area
    drag_force = -drag_magnitude * momentum.unit()
    return gravitational_force + drag_force


def tangent(t: float, state):
    position, momentum = state
    dposition_dt = momentum / object_mass
    dmomentum_dt = force(position, momentum)
    return (dposition_dt, dmomentum_dt)


time = np.linspace(0, 4, 100) * u.s
initial_position = vector.obj(
    x=0 * u.m,
    y=0 * u.m,
)
initial_momentum = (
    vector.obj(
        x=1 * u.m / u.s,
        y=20 * u.m / u.s,
    )
    * object_mass
)

position, momentum = wrapped_solve(
    fun=tangent,
    t_span=(time[0], time[-1]),
    y0=(initial_position, initial_momentum),
    t_eval=time,
)

fig, ax = plt.subplots()
ax.plot(position.x / u.m, position.y / u.m, label="with air resistance")

position_no_drag = (
    initial_momentum / object_mass * time + 0.5 * gravitational_acceleration * time**2
)

ax.plot(
    position_no_drag.x / u.m,
    position_no_drag.y / u.m,
    ls="--",
    label="no air resistance",
)

ax.set_xlabel("x (m)")
ax.set_ylabel("y (m)")
ax.set_title("Projectile motion")
ax.legend()

fig.savefig("demo.png")
```
producing
demo

## Checklist

## Before Merging

### DIFF
--- a/src/vector/__init__.py
+++ b/src/vector/__init__.py
@@ -7,6 +7,7 @@ from __future__ import annotations
 
 import importlib.metadata
 import typing
+from types import ModuleType
 
 import packaging.version
 
@@ -96,6 +97,7 @@ else:
     )
 
 
+pytree: ModuleType | None = None
 try:
     import optree.pytree
 
@@ -104,7 +106,7 @@ try:
     pytree = optree.pytree.reexport(namespace="vector")
     _register(pytree)
 except ImportError:
-    pytree = None
+    pass
 
 __all__: tuple[str, ...] = (
     "Array",
--- a/src/vector/__init__.py
+++ b/src/vector/__init__.py
@@ -96,6 +96,16 @@ else:
     )
 
 
+try:
+    import optree.pytree
+
+    from vector._pytree import _register
+
+    pytree = optree.pytree.reexport(namespace="vector")
+    _register(pytree)
+except ImportError:
+    pytree = None
+
 __all__: tuple[str, ...] = (
     "Array",
     "Azimuthal",
@@ -146,6 +156,7 @@ __all__: tuple[str, ...] = (
     "awkward_transform",
     "dim",
     "obj",
+    "pytree",
     "register_awkward",
     "register_numba",
     "zip",
--- a/src/vector/_pytree.py
+++ b/src/vector/_pytree.py
@@ -7,6 +7,7 @@ See https://blog.scientific-python.org/pytrees/ for the rationale for these func
 from __future__ import annotations
 
 from types import ModuleType
+from typing import Any
 
 from vector._methods import (
     Vector2D,
@@ -22,25 +23,28 @@ from vector.backends.object import (
     VectorObject4D,
 )
 
+Children = tuple[Any, ...]
+MetaData = tuple[type, ...]
 
-def flatten2D(v: Vector2D) -> tuple[tuple, tuple]:
+
+def flatten2D(v: Vector2D) -> tuple[Children, MetaData]:
     children = v.azimuthal.elements
     metadata = type(v), type(v.azimuthal)
     return children, metadata
 
 
-def unflatten2D(metadata: tuple, children: tuple) -> Vector2D:
+def unflatten2D(metadata: MetaData, children: Children) -> Vector2D:
     backend, azimuthal = metadata
     return backend(azimuthal=azimuthal(*children))
 
 
-def flatten3D(v: Vector3D) -> tuple[tuple, tuple]:
+def flatten3D(v: Vector3D) -> tuple[Children, MetaData]:
     children = v.azimuthal.elements, v.longitudinal.elements
     metadata = type(v), type(v.azimuthal), type(v.longitudinal)
     return children, metadata
 
 
-def unflatten3D(metadata: tuple, children: tuple) -> Vector3D:
+def unflatten3D(metadata: MetaData, children: Children) -> Vector3D:
     coords_azimuthal, coords_longitudinal = children
     backend, azimuthal, longitudinal = metadata
     return backend(
@@ -49,7 +53,7 @@ def unflatten3D(metadata: tuple, children: tuple) -> Vector3D:
     )
 
 
-def flatten4D(v: Vector4D) -> tuple[tuple, tuple]:
+def flatten4D(v: Vector4D) -> tuple[Children, MetaData]:
     children = (
         v.azimuthal.elements,
         v.longitudinal.elements,
@@ -59,7 +63,7 @@ def flatten4D(v: Vector4D) -> tuple[tuple, tuple]:
     return children, metadata
 
 
-def unflatten4D(metadata: tuple, children: tuple) -> Vector4D:
+def unflatten4D(metadata: MetaData, children: Children) -> Vector4D:
     coords_azimuthal, coords_longitudinal, coords_temporal = children
     backend, azimuthal, longitudinal, temporal = metadata
     return backend(
--- a/src/vector/_pytree.py
+++ b/src/vector/_pytree.py
@@ -1,0 +1,92 @@
+"""PyTree operations for vector objects.
+
+This module defines how vector objects are handled within optree.
+See https://blog.scientific-python.org/pytrees/ for the rationale for these functions.
+"""
+
+from __future__ import annotations
+
+from types import ModuleType
+
+from vector._methods import (
+    Vector2D,
+    Vector3D,
+    Vector4D,
+)
+from vector.backends.object import (
+    MomentumObject2D,
+    MomentumObject3D,
+    MomentumObject4D,
+    VectorObject2D,
+    VectorObject3D,
+    VectorObject4D,
+)
+
+
+def flatten2D(v: Vector2D) -> tuple[tuple, tuple]:
+    children = v.azimuthal.elements
+    metadata = type(v), type(v.azimuthal)
+    return children, metadata
+
+
+def unflatten2D(metadata: tuple, children: tuple) -> Vector2D:
+    backend, azimuthal = metadata
+    return backend(azimuthal=azimuthal(*children))
+
+
+def flatten3D(v: Vector3D) -> tuple[tuple, tuple]:
+    children = v.azimuthal.elements, v.longitudinal.elements
+    metadata = type(v), type(v.azimuthal), type(v.longitudinal)
+    return children, metadata
+
+
+def unflatten3D(metadata: tuple, children: tuple) -> Vector3D:
+    coords_azimuthal, coords_longitudinal = children
+    backend, azimuthal, longitudinal = metadata
+    return backend(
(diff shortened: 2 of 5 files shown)


src/vector/__init__.py
```diff
@@ -96,6 +97,17 @@ def _import_awkward() -> None:
     )
 
 
+pytree: ModuleType | None = None
```
nsmith-, line 100: The problem is we would like to find some way to tell the user that `optree` needs to be installed only if they attempt to access the `vector.pytree` module. We'd be better off just making a `src/vector/pytree.py` module so it can raise if not available, but then we'd have to access using `from vector.pytree import pytree`. Maybe that's not so bad?

src/vector/_pytree.py
```diff
@@ -0,0 +1,96 @@
...
+    Vector4D,
+)
+from vector.backends.object import (
+    MomentumObject2D,
+    MomentumObject3D,
+    MomentumObject4D,
+    VectorObject2D,
+    VectorObject3D,
+    VectorObject4D,
+)
```
nsmith-, line 24: I *think* we can also add numpy backends, as long as we also register the numpy array as a pytree node.

nsmith-:
FYI @matthewfeickert @pfackeldey

pfackeldey:
That's cool @nsmith-! This is a nice application for pytrees where you want to make use of the vector objects for physics, but are restricted by solvers that only work with floats.

We definitely should add this for NumPy as well - it should be rather straightforward as they are automatically considered leaf types. Awkward Arrays are much more complicated, I'm not sure if that should be supported for them, and if so how they should be decomposed.

src/vector/__init__.py
nsmith-: How about
```python
import vector

pytree = vector.register_pytree()
```
?

nsmith-: Another advantage is then the docstring for this function would be a good place to describe what it is for.

Though, a module docstring for `vector.pytree` could also work

nsmith-:
Ok so it is making a distinction between `flatten` and `ravel`, where the latter also concatenates numpy arrays. It seems then that one would have to use something like https://github.com/metaopt/optree/blob/e0e772ea3bb4ab551cb6f102b8074022bb8542ec/optree/integrations/numpy.py#L133 to actually get things fully raveled?
We could add a helper function to the re-exported namespace, `ravel` for that purpose?

nsmith-:
> We're storing things as SoA in vector right?

Well, it's a bit messy, they can be either:
```python
vec_soa = vector.obj(x=1, y=1) * np.ones(10)
print(pytree.flatten(vec_soa))
vec_aos = vector.array(dict(x=np.ones(10), y=np.ones(10)))
print(pytree.flatten(vec_aos))
```
gives
```
([array([1., 1., 1., 1., 1., 1., 1., 1., 1., 1.]), array([1., 1., 1., 1., 1., 1., 1., 1., 1., 1.])], PyTreeSpec(CustomTreeNode(VectorObject2D[(<class 'vector.backends.object.VectorObject2D'>, <class 'vector.backends.object.AzimuthalObjectXY'>)], [*, *]), namespace='vector'))
([VectorNumpy2D([(1., 1.), (1., 1.), (1., 1.), (1., 1.), (1., 1.), (1., 1.),
               (1., 1.), (1., 1.), (1., 1.), (1., 1.)],
              dtype=[('x', '<f8'), ('y', '<f8')])], PyTreeSpec(*))
```

(note `vector.obj(x=np.ones(10), y=np.ones(10))` raises, which makes sense. But the object backend is "leaky"

pfackeldey:
> Ok so it is making a distinction between `flatten` and `ravel`, where the latter also concatenates numpy arrays. It seems then that one would have to use something like https://github.com/metaopt/optree/blob/e0e772ea3bb4ab551cb6f102b8074022bb8542ec/optree/integrations/numpy.py#L133 to actually get things fully raveled? We could add a helper function to the re-exported namespace, `ravel` for that purpose?

I'm not sure what the existing ravel in optree is missing that we want to add here? Shouldn't that be enough once the tree decomposition works?

nsmith-:
Yes, so
```python
from optree.integrations.numpy import ravel_pytree

ravel_pytree(vec_soa, namespace="vector")
```
works, but I'd rather have `pytree.ravel` as a shortcut

nsmith-:
For AoS arrays, we may have to ask optree to implement something, as
```python
from optree.integrations.numpy import ravel_pytree

data = np.array(
    [
        (1.0, 2.0),
        (3.0, 4.0),
        (5.0, 6.0),
    ],
    dtype=[("x", "<f8"), ("y", "<f8")],
)

flat, unravel_func = ravel_pytree(data)
```
returns
```
(array([(1., 1.), (1., 1.), (1., 1.)], dtype=[('x', '<f8'), ('y', '<f8')]),
 functools.partial(<function _tree_unravel at 0x12618f9c0>, PyTreeSpec(*), functools.partial(<function _unravel_leaves_single_dtype at 0x12618fb00>, (3,), ((3,),))))
```
which isn't friendly to the scipy methods:
```python
from scipy.optimize import minimize

# works
minimize(lambda d: np.hypot(d[::2], d[1::2]).sum(), data.view(np.float64))

# does not work
minimize(lambda d: np.hypot(d['x'], d['y']).sum(), data)
```

src/vector/_pytree.py
```diff
@@ -0,0 +1,176 @@
...
+    )
+
+
+def _flattenAoSdata(v: VectorNumpy) -> tuple[Children, tuple[type, numpy.dtype]]:
+    assert v.dtype.fields is not None
+    field_dtypes = [dt for dt, *_ in v.dtype.fields.values()]
+    target_dtype = field_dtypes[0]
+    if not all(fd == target_dtype for fd in field_dtypes):
+        raise ValueError("All fields must have the same dtype to flatten")
+    array = numpy.array(v).view(target_dtype)
```
nsmith-, line 96: We manually flatten the numpy struct dtype here and include the necessary information to revert it. For safety we require all fields to have the same dtype. This is necessary to fully flatten the array to a 1d vector

src/vector/_pytree.py
```diff
@@ -0,0 +1,202 @@
...
+    MomentumObject2D,
+    MomentumObject3D,
+    MomentumObject4D,
+    VectorObject2D,
+    VectorObject3D,
+    VectorObject4D,
+)
+
+Children = tuple[Any, ...]
+MetaData = tuple[type, ...]
```
Saransh-cpp, line 35: FYI: the Python 3.8 job is failing because this syntax is not supported on Python 3.8. Maybe we can use a Python 3.8 friendly syntax for this PR and worry about removing 3.8 support in another issue/PR.

Saransh-cpp (commented):
TIL: optree

This is quite an amazing addition. Thanks for working on this, @nsmith-! I am a bit on the fence with the API design.

The `register_awkward` functions in Vector returns nothing and lets users pass Vector behavior names to the backend library (`awkward`) directly (`vector.register_awkward(); ak.Array(..., with_name="Momentum3D")`).

It would be nice to have an identical API for pytrees. From my very short experiments, an `awkward`-like API for `optree` looks feasible -

```py
def _flatten2D(v: Vector2D) -> tuple[Children, MetaData]:
    children = v.azimuthal.elements
    metadata = type(v), type(v.azimuthal)
    return children, metadata

def _unflatten2D(metadata: MetaData, children: Children) -> Vector2D:
    backend, azimuthal = metadata
    return backend(azimuthal=azimuthal(*children))

optree.register_pytree_node(
    VectorObject2D,
    flatten_func=_flatten2D,
    unflatten_func=_unflatten2D,
    namespace="vector.pytree"
)
```

and then the user can -
```py
leaves, treedef = optree.tree_flatten(vec, namespace="vector.pytree")
```

akin to how the `awkward` backend works (`ak.Array(..., with_name="Momentum3D")`).

We can then have a vector specific constructor (like `vector.Array` and `vector.zip`), which I think in this case is the return value of `vector.register_pytree`.

Could you see if this is feasible? Looking at how the `awkward` backend is structured will be very useful. Thanks again for working on this!

Saransh-cpp (changes requested):
Thanks, @nsmith-! Could you add that way of using optree for Vector classes in the example? I have a few other structural suggestions for the docs:
- The example should be a notebook, so that it can be tested on our CI. The notebook is definitely a tutorial so it should be moved under `Tutorials`. Can the API documentation be moved somewhere else? Maybe a new page under `Vector Constructors`?
-  [Vector backends in README](https://github.com/scikit-hep/vector?tab=readme-ov-file#backends) should be updated (perhaps with a new heading below backends?)
- The [documentation tree in README](https://github.com/scikit-hep/vector?tab=readme-ov-file#documentation) should be updated
- Same updates for [docs/index.md](https://github.com/scikit-hep/vector/blob/main/docs/index.md)

Saransh-cpp (approved):
Thanks for the updates, @nsmith-! This looks good to me now 🚀 

@henryiii @pfackeldey please feel free to merge once you approve.

nsmith-:
@henryiii or @Saransh-cpp do you think we can mint a release with this?

Saransh-cpp:
Yes! This was on my radar. Doing it right now.

PR #638, opened 2025-09-16
chore: dependency groups

henryiii:
This switches to using dependency groups, and modernizes the noxfile.

This enables `uv run pytest` to work out-of-the-box (also added the `uv.lock` file to the ignored files).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -78,7 +78,6 @@ test-extras = [
   "jax",
   "dask_awkward",
   "spark-parser",
-  'uncompyle6; python_version == "3.8"',
 ]
 docs = [
   "awkward>=2",


pyproject.toml
```diff
@@ -45,22 +45,42 @@ dependencies = [
...
+test-optional = [
+  "awkward>=2",
+  "numba>=0.57; python_version<'3.14'",
   "sympy",
+]
+test-extras = [
   "jax",
+  "dask_awkward",
+  "spark-parser",
+  'uncompyle6; python_version == "3.8"',
```
Saransh-cpp, line 81: Don't need this now?
```suggestion
```

Saransh-cpp (approved):
Thanks, @henryiii! Looks good to me!

PR #653, opened 2025-10-21
chore: rm support for Python 3.9 + bump Numba

### LINKED ISSUE #600: Replace custom `nan_to_num` with Numba's implementation
Vector has a custom Numba implementation of `numpy.nan_to_num` -

https://github.com/scikit-hep/vector/blob/6ba5a60ac90e5afa047f53681257b93ffa158cb5/src/vector/backends/_numba_object.py#L54-L55

which should ideally by in Numba.

The feature request was raised by Jim in https://github.com/numba/numba/issues/6857 and the function was partially implemented in https://github.com/numba/numba/pull/8623. I have opened a new issue for a complete implementation - https://github.com/numba/numba/issues/10089 - and we should be ready to remove our custom implementation once this issue is closed.

### LINKED ISSUE #641: Remove Python 3.8 and 3.9 support
Python 3.8 has already been EOL for a while now and 3.9 will reach EOL this year. Moreover, I contributed a vector-numba (#600) functionality upstream to numba, but we cannot get rid of the local implementation and use numba's implementation as numba now only supports Python 3.10+. Hence, adoption of a new upstream feature is being blocked by vector's support of old Python versions, which is definitely a concern. 

Blocked by #531

Saransh-cpp:
## Description

Fixes #600 
Fixes #641

Removes Python 3.9 support:
* be explicit about zip being strict
* update typing syntax
* type ignore numpy ufunc calls for object and sympy vectors
* fix getitem signature for numpy vectors and coordinate classes
* explicit type ignores for momentum numpy classes

Bumps the minimum Numba version (supported from Python 3.9):
* remove local `nan_to_num` definition

## Checklist

## Before Merging

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,6 +42,8 @@ jobs:
           - "3.11"
           - "3.12"
           - "3.13"
+          - "3.14"
+          - "3.14t"
     name: Python ${{ matrix.python-version }} - Lite
     steps:
       - uses: actions/checkout@v6
@@ -67,6 +69,8 @@ jobs:
           - "3.11"
           - "3.12"
           - "3.13"
+          - "3.14"
+          - "3.14t"
     name: Python ${{ matrix.python-version }} - Full
     steps:
       - uses: actions/checkout@v6
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,7 +19,7 @@ repos:
       - id: trailing-whitespace
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: "v0.14.10"
+    rev: "v0.14.11"
     hooks:
       - id: ruff-check
         args: ["--fix", "--show-fixes"]
--- a/noxfile.py
+++ b/noxfile.py
@@ -6,6 +6,9 @@
 
 from __future__ import annotations
 
+import os
+import sys
+import sysconfig
 from pathlib import Path
 
 import nox
@@ -16,6 +19,7 @@ nox.options.default_venv_backend = "uv|virtualenv"
 DIR = Path(__file__).parent.resolve()
 PYPROJECT = nox.project.load_toml(DIR / "pyproject.toml")
 ALL_PYTHON = nox.project.python_versions(PYPROJECT)
+ALL_PYTHON += ["3.14t"]  # add free-threaded Python variant
 
 
 @nox.session(reuse_venv=True)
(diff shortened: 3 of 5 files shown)


src/vector/backends/_numba_object.py
```diff
@@ -50,36 +50,6 @@
     _coord_object_type,
 )
 
-
-@numba.extending.overload(numpy.nan_to_num)  # FIXME: This needs to go into Numba!
-def nan_to_num(x, copy=True, nan=0.0, posinf=None, neginf=None):
```
Saransh-cpp: I added this to Numba a while back

src/vector/backends/numpy.py
```diff
@@ -191,14 +191,24 @@ def _setitem(
...
+    | AzimuthalNumpyXY
+    | LongitudinalNumpyZ
+    | LongitudinalNumpyTheta
+    | LongitudinalNumpyEta
+    | TemporalNumpyT
+    | TemporalNumpyTau,
     where: typing.Any,
     is_momentum: bool,
-) -> float | FloatArray:
+) -> float | FloatArray | VectorProtocol | Azimuthal | Longitudinal | Temporal:
```
Saransh-cpp: This also implements `__getitem__` for NumPy coordinate types. Updating the type signature allowed me to remove a few `type: ignore[arg-type]` comments in the function body.

src/vector/backends/numpy.py
```diff
@@ -398,9 +408,18 @@ class GetItem:
...
+
+    @typing.overload
+    def __getitem__(
+        self, where: typing.Any
+    ) -> float | FloatArray | VectorProtocol | Azimuthal | Longitudinal | Temporal: ...
 
-    def __getitem__(self, where: typing.Any) -> float | FloatArray:
+    def __getitem__(
+        self, where: typing.Any
+    ) -> float | FloatArray | VectorProtocol | Azimuthal | Longitudinal | Temporal:
```
Saransh-cpp: Updated overloads as per the updates to the actual function (see above)

src/vector/backends/numpy.py
```diff
@@ -1311,7 +1332,7 @@ def __setitem__(self, where: typing.Any, what: typing.Any) -> None:
         return _setitem(self, where, what, False)
 
 
-class MomentumNumpy2D(PlanarMomentum, VectorNumpy2D):
+class MomentumNumpy2D(PlanarMomentum, VectorNumpy2D):  # type: ignore[misc]
```
Saransh-cpp: The `type: ignore[misc]` comments were already present in `VectorNumpy*D` definitions, but now they are also required for `Momentum` classes.

Saransh-cpp:
@henryiii, @pfackeldey, could you please take a look whenever you are free (after the holidays 🎄)? Thanks!

ikrommyd:
You would need to change a test for 3.14 support. I'm attempting to run the vector tests with 3.14 here https://github.com/scikit-hep/integration-tests/pull/14 and I'm hitting this.
```
     def test_VectorObjectType():
        # These tests verify that the reference counts for Python objects touched in
        # the lowered Numba code do not increase or decrease with the number of times
        # the function is run.
    
        @numba.njit
        def zero(obj):
            return None
    
        @numba.njit
        def one(obj):
            return obj
    
        @numba.njit
        def two(obj):
            return obj, obj
    
        obj = vector.obj(x=1, y=2)
>       assert (sys.getrefcount(obj), sys.getrefcount(obj.azimuthal)) == (2, 2)
E       assert (1, 2) == (2, 2)
E         
E         At index 0 diff: 1 != 2
E         
E         Full diff:
E           (
E         -     2,
E         ?     ^
E         +     1,
E         ?     ^
E               2,
E           )
```

PR #659, opened 2025-10-31
feat: validate coordinate names in one place, for every backend

ikrommyd:

## Description

Rebased on `main` and reimplemented. Solves the half of #660 that vector can fix on its own.

### The problem

Each backend had its own rules for which coordinate names make a vector — `obj()`, the `VectorObject*D` constructors, `array()`, the six `__array_finalize__`, `_check_names`, the `VectorSympy*D` constructors, and the Numba `vector.obj` — and they disagreed. All of these work on `main` and silently build the wrong vector:

```python
vector.array({"pt": ..., "phi": ..., "eta": ..., "mass": ..., "energy": ...})  # energy wins, mass is dead
vector.array({"x": ..., "y": ..., "rho": ..., "phi": ...})                     # rho, phi are dead
vector.array({"x": ..., "y": ..., "z": ..., "eta": ...})                       # eta is dead
vector.zip({"rho": ..., "phi": ..., "pt": ...})                                # pt is dead
vector.VectorObject2D(x=1, px=2, y=3)                                          # x == 2
```

### The fix

`_methods._check_coordinate_names` is now the only implementation. It returns `(is_momentum, dimension, coordinates, extra)` or raises, and takes the dimension and momentum-ness a constructor is restricted to, so it serves both `vector.obj` (deduces both) and `MomentumNumpy4D` (fixes both). Every backend calls it, and the error messages come from one table instead of the nine hand-written copies. Source diff: **−1016/+567**.

The awkward behaviors validate through `__awkward_validation__` (awkward#3710, needs `awkward>=2.8.11`; older versions skip it), so what never went through a vector constructor is caught too:

```python
>>> ak.zip({"pt": pt, "eta": eta, "phi": phi, "mass": mass, "energy": energy}, with_name="Momentum4D")
TypeError: MomentumArray4D with fields [...]: specify t= or tau=, but not more than one

>>> jets["rho"] = rho
TypeError: MomentumArray4D with fields [...]: 'pt' and 'rho' both map to 'rho'
```

Subclassed behaviors (coffea's) inherit it. Cost is ~0.5 µs per array creation against ~130 µs for one `v.pt` on 100k entries; `_check_coordinate_names` is `lru_cache`d on the field-name tuple.

### Intended breaks

1. A field that is a coordinate under another name is no longer carried as payload. Ordinary extra fields (`charge`, `btag`) are unaffected.
2. Generic vectors reject momentum-aliases: `VectorObject2D(px=1, py=2)` used to return a *non*-momentum vector while `vector.obj(px=1, py=2)` returns `MomentumObject2D`. Momentum classes still take generic names.
3. An array whose fields stop describing its record name is rejected — `del jets["mass"]`, or a column-projected form that kept `__record__` but dropped coordinates. On `main` those gave a half-broken vector that failed at first coordinate access.
4. Error messages changed. No test asserted on their text.

### Tests

`tests/test_pr_659.py`: all 511 subsets of the generic names and all 222 combinations with momentum-aliases, per backend, against a reference implementation written independently of the one under test; plus the record names accepted and rejected by `__awkward_validation__`, and the numba path. 3969 tests in ~5 s, 100% coverage of the new code.

Coffea's validation and this agree on all 48,597 combinations I compared, and coffea's suite passes against this branch unchanged (its own hooks shadow vector's, so nothing there needs to change).

## Checklist

- [x] Does your submission pass pre-commit? (`$ pre-commit run --all-files` or `$ nox -s lint`)
- [x] Does the documentation build with your changes? (`$ cd docs; make clean; make html` or `$ nox -s docs`)
- [x] Does your submission pass the doctests? (`$ pytest --doctest-plus src/vector/` or `$ nox -s doctests`)

### DIFF
--- a/docs/src/make_awkward.md
+++ b/docs/src/make_awkward.md
@@ -27,7 +27,18 @@ and thus the `abs` function computes the magnitude of each record as `sqrt(x**2 
 
 It is not necessary to install Vector's behaviors globally. They could be installed in the `vec` array only by passing `behavior=vector.backends.awkward.behavior` to the [ak.Array](https://awkward-array.org/doc/main/reference/generated/ak.Array.html) constructor.
 
-The records can contain more fields than those that specify coordinates, which can be useful for specifying properties of a particle other than its momentum. Only the coordinate names are considered when performing vector calculations. Coordinates must be numbers (not, for instance, lists of numbers). Be careful about field names that coincide with coordinates, such as `rho` (azimuthal magnitude) and `tau` (proper time).
+The records can contain more fields than those that specify coordinates, which can be useful for specifying properties of a particle other than its momentum. Only the coordinate names are considered when performing vector calculations. Coordinates must be numbers (not, for instance, lists of numbers).
+
+Field names that mean a coordinate under another name, such as `rho` (a synonym of `pt`) or `energy` in a record that already has `mass`, would silently change what a vector means, so Vector rejects them:
+
+```python
+>>> ak.Array([{"pt": 1.1, "phi": 2.2, "eta": 3.3, "mass": 4.4, "energy": 5.5}], with_name="Momentum4D")
+Traceback (most recent call last):
+    ...
+TypeError: specify t= or tau=, but not more than one
+```
+
+This runs whenever the behaviors are attached, not only in `vector.Array` and `vector.zip`, so `jets["rho"] = ...` is caught too. It is the `__awkward_validation__` hook of each behavior class and needs `awkward>=2.8.11`; older versions skip it. Subclassed behaviors (see [Advanced: subclassing Awkward-Vector behaviors](awkward.ipynb)) inherit the check and can extend it by overriding `__awkward_validation__` and calling `super().__awkward_validation__()`.
 
 The `vector.Array` function (`vector.awk` is a synonym) is an alternative to the [ak.Array](https://awkward-array.org/doc/main/reference/generated/ak.Array.html) constructor, which installs Vector's behavior in the new array (not globally in `ak.behavior`).
 
--- a/src/vector/backends/numpy.py
+++ b/src/vector/backends/numpy.py
@@ -46,6 +46,7 @@ from vector._methods import (
     Vector4D,
     VectorProtocol,
     _aztype,
+    _check_coordinate_names,
     _coordinate_class_to_names,
     _coordinate_order,
     _handler_of,
@@ -865,6 +866,46 @@ class TemporalNumpyTau(TemporalNumpy, TemporalTau, GetItem, FloatArray):  # type
         return self["tau"]
 
 
+_azimuthal_numpy_type: dict[
+    str, type[AzimuthalNumpyXY] | type[AzimuthalNumpyRhoPhi]
+] = {"x": AzimuthalNumpyXY, "rho": AzimuthalNumpyRhoPhi}
+_longitudinal_numpy_type: dict[
+    str,
+    type[LongitudinalNumpyZ]
+    | type[LongitudinalNumpyTheta]
+    | type[LongitudinalNumpyEta],
+] = {
+    "z": LongitudinalNumpyZ,
+    "theta": LongitudinalNumpyTheta,
+    "eta": LongitudinalNumpyEta,
+}
+_temporal_numpy_type: dict[str, type[TemporalNumpyT] | type[TemporalNumpyTau]] = {
+    "t": TemporalNumpyT,
+    "tau": TemporalNumpyTau,
+}
+
+
+def _check_dtype_names(
+    array: VectorNumpy2D | VectorNumpy3D | VectorNumpy4D, dimension: int
+) -> tuple[tuple[str, str], ...]:
+    """
+    Validates a structured dtype against the coordinates that ``array``'s class
+    expects, returning the ``(generic name, given name)`` pairs it found.
+    """
+    if array.dtype.names is None:
+        raise TypeError(
+            f"{type(array).__name__} must have a structured dtype containing "
+            "its coordinates as fields"
+        )
+    _, _, names, _ = _check_coordinate_names(
+        array.dtype.names,
+        dimension=dimension,
+        momentum=type(array)._IS_MOMENTUM,
+        allow_extra=True,
+    )
+    return names
+
+
 class VectorNumpy(Vector, GetItem):  # noqa: PLW1641
     """Mixin class for NumPy vectors."""
 
@@ -1193,15 +1234,8 @@ class VectorNumpy2D(VectorNumpy, Planar, Vector2D, FloatArray):  # type: ignore[
         if obj is None:
             return
 
-        if _has(self, ("x", "y")):
-            self._azimuthal_type = AzimuthalNumpyXY
-        elif _has(self, ("rho", "phi")):
-            self._azimuthal_type = AzimuthalNumpyRhoPhi
-        else:
-            raise TypeError(
-                f"{type(self).__name__} must have a structured dtype containing "
-                'fields ("x", "y") or ("rho", "phi")'
-            )
+        names = _check_dtype_names(self, 2)
+        self._azimuthal_type = _azimuthal_numpy_type[names[0][0]]
 
         _is_type_safe(self)
 
@@ -1366,19 +1400,13 @@ class MomentumNumpy2D(PlanarMomentum, VectorNumpy2D):  # type: ignore[misc]
         if obj is None:
             return
 
+        names = _check_dtype_names(self, 2)
+
         # Install a fresh dtype on ``self`` rather than mutating the dtype object,
         # which is shared with the base array and would rename the caller's fields.
         self.dtype = _momentum_to_generic_dtype(self.dtype)
 
-        if _has(self, ("x", "y")):
-            self._azimuthal_type = AzimuthalNumpyXY
-        elif _has(self, ("rho", "phi")):
-            self._azimuthal_type = AzimuthalNumpyRhoPhi
-        else:
-            raise TypeError(
-                f"{type(self).__name__} must have a structured dtype containing "
-                'fields ("x", "y") or ("rho", "phi") or ("px", "py") or ("pt", "phi")'
-            )
+        self._azimuthal_type = _azimuthal_numpy_type[names[0][0]]
 
         _is_type_safe(self)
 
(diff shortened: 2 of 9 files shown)


src/vector/backends/object.py
```diff
@@ -3211,7 +3211,7 @@ def obj(**coordinates: float) -> VectorObject:
     if "E" in coordinates:
```
Saransh-cpp: Should this also have the `and "t" not in generic_coordinates` condition?

src/vector/backends/object.py
```diff
@@ -3220,7 +3220,7 @@ def obj(**coordinates: float) -> VectorObject:
     if "M" in coordinates:
```
Saransh-cpp: Ditto, for `tau`

tests/backends/test_coordinate_validation.py
Saransh-cpp: Given that these tests are related to constructing vectors, we should move them to the appropriate test backend files. The object ones can go in:

https://github.com/scikit-hep/vector/blob/e374915390e1a45f972397de053e3b788938fb81/tests/backends/test_object.py#L14

The NumPy ones (we don't have much constructor tests for NumPy at the moment, so this is perfect 😅):

https://github.com/scikit-hep/vector/blob/e374915390e1a45f972397de053e3b788938fb81/tests/backends/test_numpy.py#L85

The Awkward ones (same as NumPy):

https://github.com/scikit-hep/vector/blob/e374915390e1a45f972397de053e3b788938fb81/tests/backends/test_awkward.py#L161

Saransh-cpp (changes requested):
Good catch, @ikrommyd! Thanks for working on this!!

Please see my comments below:

tests/backends/test_coordinate_validation.py
ikrommyd: Sure. I just thought that maybe it should be a new test because it's like 1K lines but yeah I can move them under their respective backend tests.

pfackeldey:
> I think @pfackeldey might be able to help with this - can we add constructor checks to our custom Awkward behaviors?

I'm not sure about the following:
I think you can't really change the way `__init__` works through behaviors because awkward updates it's base class _after_ the constructor has been run.

If my understanding here is wrong, you _can_ do the following:
```python
class MyBehavior:
  def __init__(self, *args, **kwargs):
    super().__init__(*args, **kwargs)
    # check
    if set(self._layout.fields) != {"pt", "eta", "phi", "mass"}:
      raise ValueError('got different fields than this behavior expected')
```

However, @ikrommyd and me talked about this and these constructor checks are indeed helpful, but do not prevent to break the behaviors. What would instead prevent this thoroughly would be:
1. Awkward Arrays are fully immutable (you can't even set/update/delete fields)
2. we disallow getattr to retrieve fields so it's clear to the user if they invoke a property or access a field
3. vector constructors are only allowed to have access to certain fixed fields

(1.) and (3.) solves this issue entirely I think, (2.) is in addition to make people aware of when they invoke the `mass` property and when they access the `mass` field.

This is of course major break in API, so I'm not sure if we can/want to do this. I'm adding this here to write down the fix that solves this issue thoroughly. 
There could be ways to enforce this immutability wrt fields per array instance. Then we can preserve the current functionality in general awkward, but vector could enforce true immutability which does not allow to update/add/del fields to arrays with vector behaviors.
About (2.) I don't think there's really a way to solve this except for a API breakage in awkward. But this _is not_ the issue, it's just hard to understand _when_ one accesses a property and when a field through dot-access.

I could add a way in awkward-array to make them truly immutable that you can define through behaviors in vector. And in theory I could also think of a way to make post init constructor checks in awkward arrays possible for behaviors. 

If you want @ikrommyd & @Saransh-cpp I can prepare a PR and you can have a look? Or what do you think?

ikrommyd:
Hi @eduardo-rodrigues, it's not trivial to catch when you do `ak.zip` though and you don't use any of the vector constructor methods. Because `ak.zip` does not know what `Momentum4D` means. It does not know or care what behaviors it carries over. More than that, we can hard-code the vector behavior checks explicitly in awkward-array sure but that's not the full story. You can subclass and define your own behaviors (coffea does that a lot). So then you can do `with_name="PtEtaPhiMCandidate"` which is something that coffea defines. So what behaviors do we check? We can't hard-code all the user-defined stuff? And `with_name`, all it does is it assigns a `__record__` parameter to the array's layout to a string. It is completely blind as to what this implies.

ikrommyd:
Did a full write-up here of the issue: https://github.com/scikit-hep/vector/discussions/660
This PR solves part of it. For further discussion, I suggest we move there to not make the PR discussion messy.

pfackeldey:
Linking my proposed solution here as well: https://github.com/scikit-hep/vector/discussions/660#discussioncomment-14868529

src/vector/backends/numpy.py
```diff
@@ -2071,6 +2071,168 @@ def __setitem__(self, where: typing.Any, what: typing.Any) -> None:
...
+def _validate_numpy_coordinates(fieldnames: tuple[str, ...]) -> None:
+    """
+    Validate coordinate field names using dimension-guard pattern.
+
+    This follows the same logic as _check_names in awkward_constructors to ensure
+    consistent validation across backends.
+
+    Raises TypeError if duplicate or conflicting coordinates are detected.
+    """
+    complaint1 = "duplicate coordinates (through momentum-aliases): " + ", ".join(
```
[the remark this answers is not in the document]
ikrommyd: is there a suggestion regarding where to put these complaint strings?

tests/backends/test_coordinate_validation.py
ikrommyd: Any final suggestion here? To me, putting 1K lines under `test_issues.py` is not a great idea. My understanding is that this file is for small individual issues. The only refactoring that I think may be good here is just splitting this 1K file into multiple files per backend. Any takes?

src/vector/backends/object.py
ikrommyd: Isn't this unnecessary? This is the first if condition where `generic_coordinates` might be populated with `t`. So `and "t" not in generic_coordinates` is a useless check no? Same goes for `tau` in your other comment below.

src/vector/backends/numpy.py
```diff
@@ -2138,6 +2300,9 @@ def array(*args: typing.Any, **kwargs: typing.Any) -> VectorNumpy:
 
     is_momentum = any(x in _repr_momentum_to_generic for x in names)
 
+    # Validate coordinates using dimension-guard pattern (same as awkward _check_names)
+    _validate_numpy_coordinates(names)
```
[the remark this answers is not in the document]
ikrommyd: I've moved calling this function inside the `__array_finalize__` of the 2d, 3d, and 4d classes. The tests pass so I'll need to play with it a bit more in a terminal in case there's something I'm not remembering here.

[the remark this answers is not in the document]
ikrommyd: Oh..now I think I remember. Look at this
https://github.com/scikit-hep/vector/blob/776f405a180f6060671072bc9106eccb60abca7e/src/vector/backends/numpy.py#L2160-L2169
It makes a decision whether it is momentum or not and which class constructor to use based on the field names. That's why I think I wanted it in the `vector.array` implementation and not in the finalize. I think it's best to validate the fields first and then decide which constructor to use. I think it's best to move it back to where it was. I have reverted the change.

ikrommyd:
@henryiii send an agent over for review. I did the rebasing and consolidation with a coder and an adverserial reviewer myself. The goal was to have the validation only in one place for all backends and to have 100% test coverage.

docs/src/make_awkward.md
```diff
@@ -27,7 +27,18 @@ and thus the `abs` function computes the magnitude of each record as `sqrt(x**2
...
+Field names that mean a coordinate under another name, such as `rho` (a synonym of `pt`) or `energy` in a record that already has `mass`, would silently change what a vector means, so Vector rejects them:
+
+```python
+>>> ak.Array([{"pt": 1.1, "phi": 2.2, "eta": 3.3, "mass": 4.4, "energy": 5.5}], with_name="Momentum4D")
+Traceback (most recent call last):
+    ...
+TypeError: MomentumArray4D with fields ['pt', 'phi', 'eta', 'mass', 'energy']: specify t= or tau=, but not more than one (got 'energy', 'mass')
+```
+
+This runs whenever the behaviors are attached, not only in `vector.Array` and `vector.zip`, so `jets["rho"] = ...` is caught too. It is the `__awkward_validation__` hook of each behavior class and needs `awkward>=2.8.11`; older versions skip it. A rejected in-place assignment is only undone by `awkward>=2.12`: before that, the `TypeError` is raised but the array keeps the offending field. Subclassed behaviors (see [Advanced: subclassing Awkward-Vector behaviors](awkward.ipynb)) inherit the check and can extend it by overriding `__awkward_validation__` and calling `super().__awkward_validation__()`.
```
Saransh-cpp, line 41: Quick comment: should we move awkward's pin to `2.8.11` or is the current one okay?

Saransh-cpp (approved):
Thanks a lot for working on this, @ikrommyd! 🚀 

This cleaned up the codebase quite well.

There is one place where the current errors diverge:

```py
import awkward as ak, vector
b = vector.backends.awkward.behavior

for fields in ({"x": 3.0, "y": 4.0, "z": 12.0}, {"x": 3.0, "y": 4.0, "t": 99.0}):
    try:
        v = ak.Array([fields], with_name="Vector2D", behavior=b)
        print(list(fields), "-> abs =", abs(v)[0])
    except TypeError as e:
        print(list(fields), "->", str(e))
```
```
['x', 'y', 'z'] -> VectorArray2D with fields ['x', 'y', 'z']: these are the coordinates of a 3D vector, not of a 2D vector, allowed combinations are:

    x= y=
    rho= phi=
['x', 'y', 't'] -> VectorArray2D with fields ['x', 'y', 't']: unrecognized combination of coordinates, allowed combinations are:

    x= y=
    rho= phi=
```

I believe it is okay; given the amount of coordinate options and backends we have, it would be difficult to pin everything down to the exact same error. Do you think we should try making them the same?

docs/src/make_awkward.md
ikrommyd: I don't think so, it will just be ignored in previous versions. We effectively just have a useless function sitting there, it's fine.

ikrommyd:
@Saransh-cpp regarding the divergene you said above, do you prefer them to be the same? In the first example it is a valid combination, just not for a 2d vector. So perhaps the user got the with_name="Vector2D" wrong and not the coordinates. That was the point of the different error really.

Saransh-cpp:
Ah I see, that does make sense. Everything looks good to me as it is then. I'll merge this and prepare a release. Thanks a lot!

ikrommyd:
@Saransh-cpp if you wanna give it a day for the integration testing CI to pick it up that would be great https://github.com/scikit-hep/integration-tests
Just to make sure this didn't break anything downstreeam

And it has to be a minor version bump I guess because with this PR previously invalid vectors will now raise for users.

Saransh-cpp:
Sounds good, thanks for the heads up!

eduardo-rodrigues:
Hello @Saransh-cpp, since you pinged me: apologies for the slow reaction but I was largely offline last week. Yes, the different comment you triggered on above makes sense. All fine by me.

As for this massive PR, I did not review it per se but read a bit in diagonal, and read the threads. Great and important work, @ikrommyd 👍!

PR #685, opened 2026-02-23
fix: get rid of cupy<14 quick fix

### LINKED ISSUE #615: Switch to `nan_to_num` (rm `where`) in the future
#595 adds a workaround for #593's original CuPy issue - https://github.com/cupy/cupy/issues/9143. The workaround includes using `where` instead of `nan_to_num` in specific compute functions.

Using `where` is okay for numerical backends, but it is an issue for the SymPy backend. The shim `_lib.where` function is not a generic replacement for `np.where`, instead, it works only for this particular case (return the second argument (first value) as we know that the "normal" values are in the second argument and the "exceptional" ones are in the third argument). Therefore, this occurrence of `where` is safe to use, but should not be used out of "context" / as a drop-in replacement for `np.where`. It would be nice to remove the occurrences of `where` from compute functions (revert #595) once the CuPy issue is fixed (so that the compute functions stay generic).

Saransh-cpp:
## Description

Fixes #615

The weird CuPy behavior was fixed in v14 by @ikrommyd (thanks!).

## Checklist

## Before Merging

### DIFF
--- a/src/vector/_compute/spatial/eta.py
+++ b/src/vector/_compute/spatial/eta.py
@@ -13,7 +13,7 @@
 from __future__ import annotations
 
 import typing
-from math import inf
+from math import inf, nan
 
 import numpy
 
@@ -30,16 +30,12 @@ from vector._methods import (
 )
 
 
-# TODO: https://github.com/scikit-hep/vector/issues/615
-# revert back to `nan_to_num` implementation once
-# https://github.com/cupy/cupy/issues/9143 is fixed.
-# `lib.where` works but there is no SymPy equivalent for the function.
 def xy_z(lib, x, y, z):
-    return (
-        lib.where(
-            z != 0, lib.arcsinh(lib.where(z != 0, z / lib.sqrt(x**2 + y**2), z)), z
-        )
-        * 1
+    return lib.nan_to_num(
+        lib.arcsinh(z / lib.sqrt(x**2 + y**2)),
+        nan=lib.nan_to_num((z != 0) * inf, posinf=nan),
+        posinf=inf,
+        neginf=-inf,
     )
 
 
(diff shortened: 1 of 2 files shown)


Saransh-cpp:
Thanks for testing this locally! I haven't tested it as I just assumed it will work (and I don't want to pull out my Windows/Linux machine 😭).

This should hit the modified `xy_z` function:
```py
vec = vector.zip(
    {
        "x" : ak.Array([1],  backend="cuda"),
        "y" : ak.Array([2],  backend="cuda"),
        "z" : ak.Array([3],  backend="cuda"),
    },
)
vec.eta
```
and this should hit the other one:
```py
vec = vector.zip(
    {
        "rho" : ak.Array([1], backend="cuda"),
        "phi" : ak.Array([2], backend="cuda"),
        "z" : ak.Array([3], backend="cuda"),
    },
)
vec.eta
```

ikrommyd:
There you go. All good. It needs to be more than one element actually because if it's one element vector it treats like a scalar so it works with cupy < 14 too.
```py
In [1]: import vector

In [2]: import awkward as ak

In [3]: vec = vector.zip(
   ...:     {
   ...:         "x" : ak.Array([1,2,3],  backend="cuda"),
   ...:         "y" : ak.Array([2,3,4],  backend="cuda"),
   ...:         "z" : ak.Array([3,4,5],  backend="cuda"),
   ...:     },
   ...: )
   ...: vec.eta
Out[3]: <Array [1.1035868415601453, ..., 0.8813735870195429] type='3 * float64'>

In [4]: vec = vector.zip(
   ...:     {
   ...:         "rho" : ak.Array([1,2,3], backend="cuda"),
   ...:         "phi" : ak.Array([2,3,4], backend="cuda"),
   ...:         "z" : ak.Array([3,4,5], backend="cuda"),
   ...:     },
   ...: )
   ...: vec.eta
Out[4]: <Array [1.8184464592320668, ..., 1.2837956627431926] type='3 * float64'>
```
both of these give me `ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()` with cupy 13.6 as expected.

ikrommyd (approved):

PR #696, opened 2026-04-13
ci: enable numba testing in 3.14t

ikrommyd:
Numba now has 3.14t wheels. We should enable testing.
Partially reverts the dependency split done in https://github.com/scikit-hep/vector/pull/676 for the exact reason that numba did not have wheels at the time.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -80,9 +80,7 @@ def doctests(session: nox.Session) -> None:
 @nox.session(reuse_venv=True, default=False)
 def notebooks(session: nox.Session) -> None:
     """Run the notebook tests"""
-    test_deps = nox.project.dependency_groups(
-        PYPROJECT, "test", "test-optional", "test-numba"
-    )
+    test_deps = nox.project.dependency_groups(PYPROJECT, "test", "test-optional")
     session.install("-e.", *test_deps)
     session.install("jupyter")
     session.run("pytest", "tests/test_notebooks.py", *session.posargs)
--- a/noxfile.py
+++ b/noxfile.py
@@ -52,9 +52,7 @@ def tests(session: nox.Session) -> None:
         sysconfig.get_config_var("Py_GIL_DISABLED")
     ):
         os.environ["PYTHON_GIL"] = "0"
-        test_deps = nox.project.dependency_groups(PYPROJECT, "test-all-no-gil")
-    else:
-        test_deps = nox.project.dependency_groups(PYPROJECT, "test-all-gil")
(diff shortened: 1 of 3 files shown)


tests/backends/test_numba_object.py
```diff
@@ -68,7 +68,10 @@ def two(obj):
                 vector.backends.object.VectorObject2D
             )
         else:
-            assert class_refs == sys.getrefcount(vector.backends.object.VectorObject2D)
+            assert (
+                abs(class_refs - sys.getrefcount(vector.backends.object.VectorObject2D))
+                <= 1
+            )
```
ikrommyd, line 74: The refcount fluctuates between +- 1 and is different in each step of the loop locally. I thought just bounding it is fine to make the test pass. Same for the other two cases below.

pyproject.toml
Saransh-cpp, line 75: We should move numba back to `test-optional` and remove `test_numba` from `test-all`

noxfile.py
Saransh-cpp: `test-numba` won't be required here then

tests/backends/test_numba_object.py
Saransh-cpp: Do we know why this is happening? Is there something in Numba's changelog?

Saransh-cpp (changes requested):
Thanks a lot for looking into this, @ikrommyd! Please see my comments below:

pyproject.toml
ikrommyd: Ah I just kept it as separate in case we wanna split it in the future again. You want to completely undo the split of your previous PR. On it.

tests/backends/test_numba_object.py
ikrommyd: Honestly I have no idea and I tried investigating very briefly without finding anything out. Then I had claude take a shot at it to make the test pass without adding the +-1 bound for an hour without a result. It was doing trial n error but the += 1 looked random so it couldn't find a predicable assertion. I can take a more detailed look in the numba changes I guess.

ianna: > Do we know why this is happening? Is there something in Numba's changelog?

Python 3.14t has broken the assumption that reference counts are deterministic and immediate.

pyproject.toml
ikrommyd: this should be good now

tests/backends/test_numba_object.py
Saransh-cpp: Thanks, both!

Saransh-cpp (approved):
Thanks, @ikrommyd!

PR #705, opened 2026-05-21
fix: _wrap_results skips momentum equivalents

### LINKED ISSUE #704: `neg3D` gives wrong results when paired with `register_awkward`
### Vector Version

1.8.0

### Python Version

3.14

### OS / Environment

OSX with uv

### Describe the bug

Pointed out in: https://github.com/scikit-hep/coffea/issues/1565#issuecomment-4503464169

The slimmest MWE:

```py
import awkward as ak
import vector
vector.register_awkward()


vec_ak = ak.zip({"pt": [1, 2, 3, 4], "eta": [1, 2, 3, 4], "phi": [1, 2, 3, 4], "mass": [1, 2, 3, 4]}, with_name="Momentum4D")
vec_vec = vector.zip({"pt": [1, 2, 3, 4], "eta": [1, 2, 3, 4], "phi": [1, 2, 3, 4], "mass": [1, 2, 3, 4]})

print(type(vec_ak.neg3D), type(vec_vec.neg3D))
# <class 'vector.backends.awkward.MomentumArray3D'> <class 'vector.backends.awkward.MomentumArray4D'>
```

The result is being wrapped wrong.

### Any additional but relevant log output

_No response_

Saransh-cpp:
## Description

Fixes #704

More generally, `_wrap_results` skipped momentum coordinates at some places. Vector internally converts coordinates so everything works (and worked), but using an Awkward Array with Vector behaviors did not work (as there is no `t` or `tau` in fields, but there is `m` for instance).

## Checklist

## Before Merging

### DIFF
--- a/src/vector/backends/awkward.py
+++ b/src/vector/backends/awkward.py
@@ -720,7 +720,9 @@ class VectorAwkward:
                         names.append(name)
                         arrays.append(self[name])
 
-            if any(f in fields for f in ("t", "tau", "M", "m", "mass", "E", "e", "energy")):
+            if any(
+                f in fields for f in ("t", "tau", "M", "m", "mass", "E", "e", "energy")
+            ):
                 cls = cls.ProjectionClass4D
             elif any(f in fields for f in ("z", "pz", "theta", "eta")):
                 cls = cls.ProjectionClass3D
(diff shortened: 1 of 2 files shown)


ikrommyd (approved):
Works with the reproducer. Thanks!

Is it worth doing a scan (probably with an LLM) for similar issues in other functions. We found it via `boostCM_of` but maybe there's more?

Saransh-cpp:
Thanks for the review!

The issue wasn't in a particular compute function, rather it was a "wrapping issue" (how the result of any compute function is wrapped into the relevant Vector class and returned to the user). So in a sense yes, there would be quite a few compute functions that were breaking because of this, but now they're all simultaneously fixed, if that makes sense.

testing on one compute function which we know was broken for sure looks good to me, as every other compute function will also flow through the same wrapping code. I'll merge it in! Should I also create a bug fix release?

ikrommyd:
Uhmm yeah a release should be good!

Issue #711, opened 2026-06-11
Code review: bugs, performance, simplifications, and modernizations


henryiii:
:robot: _AI text below_ :robot:

A multi-agent code review of the full codebase (core, compute layer, all five backends, infrastructure). Every high-severity finding was reproduced by execution before being listed. Draft PRs addressing groups of these findings are linked below.

## High-severity bugs (all reproduced)

- [ ] **`!=` / `not_equal` uses AND instead of OR.** `src/vector/_compute/planar/not_equal.py:35`, all `spatial/not_equal.py` signatures, `lorentz/not_equal.py:146`. `xy_xy` returns `(x1 != x2) & (y1 != y2)`; the negation of equality needs `|`. The Lorentz version is doubly wrong: `(t1 != t2) & spatial_equal(...)`. Reproduced: `vector.obj(x=1.0, y=2.0) != vector.obj(x=1.0, y=3.0)` returns `False` while `==` also returns `False`. Tests only cover identical vectors.
- [ ] **`is_perpendicular` missing absolute value.** `planar/is_perpendicular.py:56`, `spatial/is_perpendicular.py:66`. The check is `dot < tol·|v1|·|v2|`, true for any obtuse pair. Reproduced: `(1,0).is_perpendicular((-1,0))` → `True`. Needs `|dot| < …`.
- [ ] **NumPy backend `v[mask] = w` broken twice over.** `backends/numpy.py:186-190`. (a) `generic` only bound in the `if is_momentum:` branch but used unconditionally → `UnboundLocalError` for non-momentum arrays (reproduced). (b) `tofill = array[where]` is a copy for boolean/fancy indices, so writes are silently discarded — only slice assignment works.
- [ ] **`vector.array(existing_structured_array)` always returns 2D.** `backends/numpy.py:2148-2169`. Field-name detection only inspects dicts and `dtype=` kwargs; a pre-built `(x,y,z,t)` structured array silently becomes `VectorNumpy2D` (reproduced). Should fall back to `asarray(args[0]).dtype.names`.
- [ ] **NumPy coordinate classes shadow `ndarray.dtype` with mutable class state.** `numpy.py:465-472` and six siblings. `AzimuthalNumpyXY.dtype = numpy.dtype(...)` in `__new__` makes every instance report whatever dtype was constructed last, anywhere in the process. Poisons `__eq__`'s dtype check, is test-order-dependent, and is a data race under free-threaded 3.14t.
[...]

PR #712, opened 2026-06-11
chore: fix noxfile, CI pass gate, and stale config

henryiii:

## Summary

- **noxfile `docs` session**: `session.install("-e.", doc_deps)` was passing the deps list as a single argument (TypeError at runtime). Fixed to `session.install("-e.", *doc_deps)`.
- **noxfile free-threaded env**: Was inspecting `sys.version_info`/`sysconfig` of the nox host interpreter and mutating `os.environ["PYTHON_GIL"]`, leaking into subsequent sessions. Now detects free-threaded builds from the `"t"` suffix on `session.python` and passes `env={"PYTHON_GIL": "0"}` to `session.run(...)` only. Applied to both `lite` and `tests` sessions (CI runs `lite` on 3.14t). Drops unused `os`/`sys`/`sysconfig` imports.
- **CI `pass` gate**: Added `if: always()` and a step that fails when any upstream job is in `failure` or `cancelled` state, so skipped jobs no longer satisfy branch protection.
- **numba lower bound**: Aligned `test-optional` group from `>=0.57` to `>=0.62` to match the published `[numba]` extra.
- **Duplicate pylint disable**: Removed the second `"cast_python_value"` entry from `pyproject.toml`.
- **Dead `requirements-txt-fixer` hook**: Removed from `.pre-commit-config.yaml` (no `requirements*.txt` files exist in the repo).
- **`blacken-docs` black pin**: Bumped `additional_dependencies: [black~=24.0]` → `black~=25.0`; pre-commit.ci autoupdate never touches `additional_dependencies`.

Part of #711

All changes validated: `yaml.safe_load` on CI YAML OK, `uvx nox -l` lists sessions without error, `prek -a --quiet` passes clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,7 +15,6 @@ repos:
       - id: end-of-file-fixer
         exclude: ^docs
       - id: mixed-line-ending
-      - id: requirements-txt-fixer
       - id: trailing-whitespace
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
@@ -54,7 +53,7 @@ repos:
     hooks:
       - id: blacken-docs
         args: ["-E"]
-        additional_dependencies: [black~=24.0]
+        additional_dependencies: [black~=25.0]
 
   - repo: https://github.com/pre-commit/pygrep-hooks
     rev: v1.10.0
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -53,7 +53,7 @@ repos:
     hooks:
       - id: blacken-docs
         args: ["-E"]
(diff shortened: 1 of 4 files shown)


.pre-commit-config.yaml
```diff
@@ -54,7 +53,7 @@ repos:
     hooks:
       - id: blacken-docs
         args: ["-E"]
-        additional_dependencies: [black~=24.0]
+        additional_dependencies: [black~=25.0]
```
henryiii:
```suggestion
        additional_dependencies: [black~=26.0]
```

PR #733, opened 2026-07-24
ci: run Awkward backend tests on GPU

henryiii:

Mirrors the GPU CI setup in [awkward](https://github.com/scikit-hep/awkward/blob/main/.github/workflows/property-tests-gpu.yml), which runs on the scikit-hep self-hosted runners we should also have access to.

- `.github/workflows/gpu.yml` — nightly, push to `main`, and `workflow_dispatch`; matrixed over CUDA 12 and 13. Micromamba env with `cupy`, then run the CUDA tests. Not triggered on pull requests, to keep PRs off the shared runners.
- `tests/cuda/test_awkward_cuda.py` — vector's Awkward backend already picks `lib` from the layout's nplike, so a cuda-backed array computes with `cupy`. The tests run 19 unary ops, 13 binary ops, and all 12 4D coordinate round-trips on both backends and require the results to match, plus a check that results stay on the `cuda` backend.
- Registered the `cuda` pytest marker.

The test module `importorskip`s cupy and skips when no device is present, so the regular CI jobs collect and skip it.

Draft because the GPU job has not run yet — it needs a `workflow_dispatch` from this branch to confirm the runners pick it up. One thing to watch there: `filterwarnings = ["error", ...]` is repo-wide, so any warning from cupy or awkward's CUDA kernel loading will fail the job.

### DIFF
--- a/.github/workflows/gpu.yml
+++ b/.github/workflows/gpu.yml
@@ -1,0 +1,66 @@
+name: GPU
+
+on:
+  schedule:
+    - cron: "19 5 * * *"
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ startsWith(github.ref, 'refs/pull/') }}
+
+env:
+  FORCE_COLOR: "1"
+
+jobs:
+  gpu-tests:
+    name: CUDA ${{ matrix.cuda-version }}
+    if: github.event_name != 'schedule' || github.repository_owner == 'scikit-hep'
+    runs-on: self-hosted
+    timeout-minutes: 60
+
+    # Required for micromamba to activate conda
+    defaults:
+      run:
+        shell: bash -l {0}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        cuda-version:
+          - 12
+          - 13
+
+    steps:
+      - name: Clean the workspace and micromamba
+        run: |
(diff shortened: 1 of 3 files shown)


henryiii:
I can't run the test from here, so it needs a merge to be tested.

ianna:
> I can't run the test from here, so it needs a merge to be tested.

would you like me to review it so that you could merge it?

ianna (approved):
@henryiii - Great! This looks consistent with awkward CUDA workflow. Thanks!

henryiii:
Okay, let's merge and then followup on failures. Once merged it can be triggered via workflow-dispatch.

henryiii:

The GPU workflow ran on the merge to `main` and passed on real GPUs: [run 30124750331](https://github.com/scikit-hep/vector/actions/runs/30124750331).

| Job | Result |
|---|---|
| CUDA 12 | **34 passed** in 5.90s |
| CUDA 13 | **34 passed** in 6.08s |

Both jobs picked up the scikit-hep self-hosted runner, so the runner access assumption held.

Two things I checked specifically, since a green tick alone would not have proved much:

- **The tests ran rather than skipped.** `collected 34 items`, all `PASSED`, zero skips. A missing cupy or a GPU-less machine would have skipped the whole module at import and still exited 0, and the job would have looked identical. It did not; cupy imported and `getDeviceCount()` found a device.
- **The warnings-as-errors concern did not materialize.** Nothing from cupy or awkward's CUDA kernel loading tripped `filterwarnings = ["error", ...]`, so no ignore entry is needed for now.

The run takes about a minute per CUDA version, so the nightly cron is cheap.

PR #729, opened 2026-07-06
Add guards for spatial conversion assignment warnings

researchingadi:
## Description

Adds targeted guards for a small subset of the Pylint use-before-assignment warnings reported in #512.

This PR focuses only on the spatial conversion helpers in:

* `src/vector/_compute/spatial/rotate_quaternion.py`
* `src/vector/_compute/spatial/cross.py`

The changes initialize conditionally assigned conversion callables before the coordinate-system branches and assert that each required callable was resolved before use. This keeps behavior unchanged for supported coordinate-system combinations while giving clearer failures if an unsupported branch reaches the conversion code.

Related issue: Part of #512

## Checklist

* [x] Have you followed the guidelines in our Contributing document?
* [x] Have you checked to ensure there aren't any other open Pull Requests for the required change?
* [ ] Does your submission pass pre-commit? (`$ pre-commit run --all-files` or `$ nox -s lint`)
* [ ] Does your submission pass tests? (`$ pytest` or `$ nox -s tests`)
* [ ] Does the documentation build with your changes? (`$ cd docs; make clean; make html` or `$ nox -s docs`)
* [ ] Does your submission pass the doctests? (`$ pytest --doctest-plus src/vector/` or `$ nox -s doctests`)

## Before Merging

* [ ] Summarize the commit messages into a brief review of the Pull request.

## AI usage disclosure

This PR was developed with substantial AI assistance. I used ChatGPT to help understand the pylint warning, draft parts of the PR/comment wording.

I personally reviewed the changed files, applied the edits, ran the targeted pylint command locally, checked the diff, and pushed the commits. The PR was not generated or submitted by an autonomous agent, but the implementation and communication were AI assisted.

PR #737, opened 2026-08-05
fix: give a useful error when ProjectionClass*D is missing

### LINKED ISSUE #488: Throw a useful error on implicit upcast if `ProjectClass*D` is not defined by a vector subclass
> Is this a way to signal that there is no 4D interpretation of VertexArray? (which is true)

> I can just remove L139, but vector would throw an error saying something like "ProjectionClass4D is not defined for VertexArray" if a user tries an operation that needs converting the array to a 4D vector. The error does not tell the user what they are doing wrong, but I am hoping that not a lot of users will try converting it to 4D?

> I think it might make sense to raise an error if a user tries to write a code that requires implicit upcast from 3D point to 4D

_Originally posted by @nsmith- in https://github.com/CoffeaTeam/coffea/pull/991#discussion_r1670580026_

FanxinSun:
Closes #488.

If you build a vector subclass from the awkward mixins and don't give it a `ProjectionClass3D`/`4D` (like coffea's `VertexArray`, which intentionally has no 4D meaning), any operation that tries to promote the result currently dies with an unhelpful `AttributeError` from awkward's field lookup, which says nothing about what's actually missing.

This PR routes the `ProjectionClass*D` lookups in the awkward backend through a small helper that raises a `TypeError` naming the class, the missing attribute, and how to fix it:

> VertexArray does not define ProjectionClass3D: this operation implicitly converts a VertexArray into a 3D vector, but no 3D equivalent has been declared for it. Assign `VertexArray.ProjectionClass3D` to a suitable vector class to allow the conversion, or convert the vectors explicitly before the operation.

Also adds a regression test that builds a minimal mixin-only `Vertex` behavior the same way coffea does. Full awkward backend test file passes locally (51 passed).

### DIFF
--- a/src/vector/backends/awkward.py
+++ b/src/vector/backends/awkward.py
@@ -589,7 +589,7 @@ def _projection_class(
 ) -> type[VectorProtocol]:
     name = f"ProjectionClass{dimension}D"
     projection: type[VectorProtocol] | None = getattr(cls, name, None)
-    if projection is None:
+    if not projection:
         msg = f"{dimension}D conversion for {cls.__name__} is not defined"
         raise TypeError(msg)
     return projection
--- a/src/vector/backends/awkward.py
+++ b/src/vector/backends/awkward.py
@@ -590,13 +590,7 @@ def _projection_class(
     name = f"ProjectionClass{dimension}D"
     projection: type[VectorProtocol] | None = getattr(cls, name, None)
     if projection is None:
-        msg = (
-            f"{cls.__name__} does not define {name}: this operation implicitly "
-            f"converts a {cls.__name__} into a {dimension}D vector, but no "
-            f"{dimension}D equivalent has been declared for it. Assign "
-            f"`{cls.__name__}.{name}` to a suitable vector class to allow the "
-            "conversion, or convert the vectors explicitly before the operation."
-        )
+        msg = f"{dimension}D conversion for {cls.__name__} is not defined"
         raise TypeError(msg)
     return projection
 
--- a/src/vector/backends/awkward.py
+++ b/src/vector/backends/awkward.py
@@ -584,6 +584,23 @@ class TemporalAwkwardTau(TemporalAwkward, TemporalTau):
         return (self.tau,)
 
 
+def _projection_class(
+    cls: type[VectorProtocol], dimension: int
+) -> type[VectorProtocol]:
+    name = f"ProjectionClass{dimension}D"
+    projection: type[VectorProtocol] | None = getattr(cls, name, None)
+    if projection is None:
+        msg = (
+            f"{cls.__name__} does not define {name}: this operation implicitly "
+            f"converts a {cls.__name__} into a {dimension}D vector, but no "
+            f"{dimension}D equivalent has been declared for it. Assign "
+            f"`{cls.__name__}.{name}` to a suitable vector class to allow the "
+            "conversion, or convert the vectors explicitly before the operation."
+        )
+        raise TypeError(msg)
+    return projection
+
+
 def _class_to_name(cls: type[VectorProtocol]) -> str:
     # respect the type of classes inheriting VectorAwkward classes
     is_vector = "vector.backends" in cls.__module__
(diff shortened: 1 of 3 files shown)


src/vector/backends/awkward.py
```diff
@@ -584,6 +584,23 @@ def elements(self) -> tuple[ArrayOrRecord]:
...
+    name = f"ProjectionClass{dimension}D"
+    projection: type[VectorProtocol] | None = getattr(cls, name, None)
+    if projection is None:
+        msg = (
+            f"{cls.__name__} does not define {name}: this operation implicitly "
+            f"converts a {cls.__name__} into a {dimension}D vector, but no "
+            f"{dimension}D equivalent has been declared for it. Assign "
+            f"`{cls.__name__}.{name}` to a suitable vector class to allow the "
+            "conversion, or convert the vectors explicitly before the operation."
+        )
```
Saransh-cpp: 5 lines of error is definitely not required:
```suggestion
        msg = (
            f"{dimension}D conversion for {cls.__name__} is not defined"
        )
```

src/vector/backends/awkward.py
```diff
@@ -584,6 +584,23 @@ def elements(self) -> tuple[ArrayOrRecord]:
         return (self.tau,)
 
 
+def _projection_class(
+    cls: type[VectorProtocol], dimension: int
+) -> type[VectorProtocol]:
+    name = f"ProjectionClass{dimension}D"
+    projection: type[VectorProtocol] | None = getattr(cls, name, None)
+    if projection is None:
```
Saransh-cpp: Nit
```suggestion
    if not projection:
```

src/vector/backends/awkward.py
```diff
@@ -584,6 +584,23 @@ def elements(self) -> tuple[ArrayOrRecord]:
         return (self.tau,)
 
 
+def _projection_class(
+    cls: type[VectorProtocol], dimension: int
+) -> type[VectorProtocol]:
+    name = f"ProjectionClass{dimension}D"
+    projection: type[VectorProtocol] | None = getattr(cls, name, None)
```
Saransh-cpp, line 591: Is the type explicitly required here? Does mypy complain?

tests/backends/test_awkward.py
```diff
@@ -1261,3 +1261,30 @@ def test_record_method_chaining_without_registration():
         check=False,
     )
     assert result.returncode == 0, result.stderr
+
+
+def test_missing_projection_class_message():
```
Saransh-cpp: This should be in test_issues. Could you please also add 1 test case each for each dimension (perhaps a for loop)? Thanks!

Saransh-cpp (approved):
Thanks a lot for working on this, @FanxinSun! As with your other PR: Please see https://github.com/scikit-hep/vector/pull/729#issuecomment-4956430952 for my take on AI generated comments / code. I'll be happy to review this PR further if there is minimal AI usage (given that it is a good first issue).

src/vector/backends/awkward.py
FanxinSun: I checked with:
```
mypy: 2.1.0
Python: 3.14.4
```
it will not complain. And this type annotation is not required, but one caveat is about warn_unreachable. Currently it's set false, by flipping it the un-annotated function will raise
`error: Statement is unreachable [unreachable]`
because ProjectionClass4D is declared non-Optional.

But in general I'm not clear about the plan of that flag so keeping it or not is up to you!

tests/backends/test_awkward.py
FanxinSun: Done. Now the test function is removed from test_akward.py and lies as test_issue_488 in test_issues.py. Looped on each dimensions and got:
`--- dimension=2  fields=['x', 'y'] ---
  VertexArray does not define ProjectionClass2D: this operation implicitly
  converts a VertexArray into a 2D vector, but no 2D equivalent has been
  declared for it. Assign `VertexArray.ProjectionClass2D` to a suitable vector
  class to allow the conversion, or convert the vectors explicitly before the
  operation.

--- dimension=3  fields=['x', 'y', 'z'] ---
  VertexArray does not define ProjectionClass3D: this operation implicitly
  converts a VertexArray into a 3D vector, but no 3D equivalent has been
  declared for it. Assign `VertexArray.ProjectionClass3D` to a suitable vector
  class to allow the conversion, or convert the vectors explicitly before the
  operation.

--- dimension=4  fields=['x', 'y', 'z', 't'] ---
  VertexArray does not define ProjectionClass4D: this operation implicitly
  converts a VertexArray into a 4D vector, but no 4D equivalent has been
  declared for it. Assign `VertexArray.ProjectionClass4D` to a suitable vector
  class to allow the conversion, or convert the vectors explicitly before the
  operation.`

PR #739, opened 2026-08-13
chore: secure CI

henryiii:
Prompted by https://github.com/pypa/hatch/issues/2292.

- Pin every GitHub Action to a commit SHA with a version comment.
- Add `permissions: {}` at the workflow level and give each job only the permissions it uses. The publish job keeps `id-token`, `attestations`, and `contents`, each with a comment.
- Add `persist-credentials: false` to all checkout steps.
- Move matrix and context values into environment variables in `run` blocks to prevent template injection.
- Add a concurrency group to the CD workflow, and names to the jobs that had none.
- Add the zizmor pre-commit hook, and freeze all hooks to SHAs.
- Update dependabot to monthly grouped updates with a 7-day cooldown, and add a pre-commit ecosystem.
- Add a 7-day cooldown for uv (`[tool.uv] exclude-newer`) and pixi (`exclude-newer` in the workspace table).

Held back at their current versions, frozen to SHAs, because the new releases make unrelated source code fail: ruff v0.15.20 (v0.16.1 reformats source), mypy v2.1.0 (v2.3.0 reports new errors), prettier v3.9.4 (v3.9.6 reformats a Markdown link). Each needs its own change set.

`hynek/build-and-inspect-python-package` moves from v2 to v3.0.1, a major version change. v3 stops force-tagging minor releases; it has no functional change beyond dependency updates.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,15 +11,3 @@ updates:
           - "*"
     cooldown:
       default-days: 7
-
-  # Maintain pre-commit hooks
-  - package-ecosystem: "pre-commit"
-    directory: "/"
-    schedule:
-      interval: "monthly"
-    groups:
-      pre-commit:
-        patterns:
-          - "*"
-    cooldown:
-      default-days: 7
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,8 +4,22 @@ updates:
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
-      interval: "weekly"
+      interval: "monthly"
     groups:
-      actions:
+      github-actions:
         patterns:
           - "*"
+    cooldown:
+      default-days: 7
+
+  # Maintain pre-commit hooks
+  - package-ecosystem: "pre-commit"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    groups:
+      pre-commit:
+        patterns:
+          - "*"
+    cooldown:
+      default-days: 7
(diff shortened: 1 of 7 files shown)


.github/dependabot.yml
```diff
@@ -4,8 +4,22 @@ updates:
...
+  - package-ecosystem: "pre-commit"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    groups:
+      pre-commit:
+        patterns:
+          - "*"
+    cooldown:
+      default-days: 7
```
henryiii:
```suggestion
```

.pre-commit-config.yaml
```diff
@@ -19,14 +19,14 @@ repos:
       - id: trailing-whitespace
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: "v0.15.20"
+    rev: "c59bba8fb259db0fec2bbb77ad8ba51ea7341b56" # frozen: v0.15.20
```
Saransh-cpp: Do we need the `"..."` here?
```suggestion
    rev: c59bba8fb259db0fec2bbb77ad8ba51ea7341b56 # frozen: v0.15.20
```

Saransh-cpp (approved):
Thanks, @henryiii! 🚀 

The failing CI is not related to this PR. I am seeing the failure in #735 too.

Edit: the `pass` check does not seem to be working?

henryiii:
pass -> Pass (renamed, needs update in settings)

.github/workflows/cd.yml
henryiii:
```suggestion
        uses: actions/attest@1e69f48acb82d1966a394da916b4c1698aa569d6 # v4.2.2
```

PR #741, opened 2026-09-07
perf(awkward): gather each operand's coordinates with one carry

lgray:

## What this fixes

Arithmetic on `ak.combinations` / `ak.cartesian` output spends most of its time inside
awkward, not inside vector — but the reason is something vector controls.

Those operations hand you an `IndexedArray` view. Awkward projects such a view **once per
field asked of it**, and each projection re-runs the `awkward_IndexedArray_getitem_nextcarry`
kernel over the whole index. A 4D vector asks for four coordinates, so `j1 + j2` ran
**eight** index kernels where two would do.

Separately, `dispatch` asks each operand for a coordinate group three times — `_aztype`
tests with `hasattr`, reads `type(obj.azimuthal)`, and then the argument list takes
`.elements` — and each rebuilds the group from scratch with `ak.fields` plus per-field
`__getitem__`. That is invisible on large arrays and dominant on small ones.

## Mechanism

Every one of the 82 `dispatch` functions has the same shape:

```python
function, *returns = _from_signature(__name__, dispatch_map, (_aztype(v1), _ltype(v1), ...))
...
handler._wrap_dispatched_function(function)(_lib_of(v1, v2), *v1.azimuthal.elements, ...)
```

The signature lookup asks each operand for exactly the coordinate groups the selected
function reads — `rho2` asks for azimuthal only, `deltaR` for azimuthal and longitudinal,
`add` for all three — and it always runs **before** `_wrap_dispatched_function`, which in
turn always runs before the argument list takes any `.elements`. So that one call is the
point where the needed field set is known and no element has been fetched yet.

- The coordinate properties route through `VectorAwkward._coordinates`, which records the
  group and reuses the built object for the rest of the dispatch (3 constructions -> 1).
- `_wrap_dispatched_function` gathers each operand once for exactly the recorded groups:
  slice the record to those fields, hoist the shared `Index` if awkward pushed it inside
  the record, and `project()`. An op reading `k` fields then pays **1 index kernel + k
  gathers** per operand, where `main` pays `k` kernels + k gathers.
- The record is thread-local and is cleared when the dispatch returns. **Nothing is stored
  on the array** — no instance cache, so no stale-cache question. If a dispatch raises
  between the wrap and its call (a `TypeError` from `_lib_of`, say) the record survives
  until the next dispatch on that thread, which drops it at its own wrap — before taking
  any elements, so a stale gather is never read and is never gathered again.

The gather is skipped entirely (falling back to `main`'s behaviour) for typetracer arrays —
so dask-awkward's column projection is untouched — and for flat, option-typed and
non-shared-index layouts, where there is no index to collapse. Those still get the
construction reuse.

Because the gather now happens in `_wrap_dispatched_function` rather than inside the
`azimuthal` property, an exception from the layout surgery can no longer be swallowed by
the `hasattr(obj, "azimuthal")` guard in `_aztype`.

### Kernel/gather counts

Per-field witness on `ak.combinations(4D vectors, 2)`, counting `IndexedArray.project`
and `NumpyArray._carry`:

| operation | fields carried | index kernels, `main` | index kernels, this PR |
|---|---|---|---|
| `u.rho2` | x, y | 2 | 1 |
| `u.phi` | x, y | 2 | 1 |
| `u.pt` | x, y | 2 | 1 |
| `u.deltaphi(v)` | x, y (both) | 4 | 2 |
| `u.deltaR(v)` | x, y, z (both) | 6 | 2 |
| `u.mass` | x, y, z, t | 4 | 1 |
| `u + v` | x, y, z, t (both) | 8 | 2 |

The carried-field lists are identical on both sides: this gathers **only** the fields the
dispatched function reads, never the whole record. Non-coordinate fields (`btag` and
friends) are never gathered on either side.

## Measurements

Real input: first 100k events of `Run2012B_SingleMu.root` (252,917 jets, 2,461,785
trijets from `ak.combinations(jets, 3)`), plus a synthetic flat 100k array and the file's
per-event MET. **Every number is a cold first access** — this PR keeps no cache between
dispatches, so there is no warm variant to quote. min-of-20 per run (min-of-100 for the
flat rows), best of 5 interleaved runs per side, single process, `awkward 2.13.0`,
CPython 3.13, macOS arm64.

| case | `main` (ms) | this PR (ms) | |
|---|---|---|---|
| trijet xyzt `j1+j2` | 26.93 | 19.76 | **-27%** |
| trijet xyzt `j1+j2+j3` | 41.77 | 30.79 | **-26%** |
| trijet xyzt `(j1+j2+j3).mass` | 46.33 | 35.02 | **-24%** |
| trijet ptetaphim `j1+j2` | 109.33 | 102.04 | -7% |
| trijet ptetaphim `j1+j2+j3` | 214.25 | 203.27 | -5% |
| trijet xyzt `j1.rho2` | 7.50 | 6.31 | **-16%** |
| trijet xyzt `j1.phi` | 13.13 | 11.97 | -9% |
| trijet xyzt `j1.pt` | 7.89 | 6.71 | **-15%** |
| trijet xyzt `j1.deltaphi(j2)` | 39.10 | 36.83 | -6% |
| trijet xyzt `j1.deltaR(j2)` | 83.41 | 78.64 | -6% |
| flat 100k xyzt `v+v` | 1.00 | 0.66 | **-34%** |
| flat 100k xyzt `v.mass` | 0.69 | 0.47 | **-31%** |
| flat 100k xyzt `v.deltaR(v)` | 10.11 | 9.68 | -4% |
| option-typed pairs `j1+j2` | 7.35 | 7.05 | -4% |
| option-typed pairs `j1.deltaR(j2)` | 17.48 | 17.09 | -2% |
| event MET `.deltaphi(jagged jets)` | 6.44 | 6.20 | -4% |
| event MET `+ jagged jets` (2D) | 3.16 | 2.96 | -6% |

No row is slower. The partial-coordinate rows (`rho2`, `phi`, `pt`, `deltaphi`) are the
ones an earlier revision of this branch regressed by gathering the whole record; gathering
only the dispatched groups turns them into wins. The `ptetaphim` rows gain least because
`pt/eta/phi/mass` arithmetic is transcendental-bound, not gather-bound. The option-typed
and event-broadcast rows are not gathered at all — their few percent is the construction
reuse alone.

### Memory

`u = tri.j1; u.rho2`, result discarded, measuring numpy bytes still reachable from `u`:

| | reachable from `u` before | after | pinned by the op |
|---|---|---|---|
| `main` | 25.55 MB | 25.55 MB | **0.00 MB** |
| this PR | 25.55 MB | 25.55 MB | **0.00 MB** |

Identical to `main`, as it must be: nothing is stored on the array. (The same probe reads
+39.39 MB against the instance-cache revision this branch previously carried, so it is a
live instrument, not an absent one.)

## What is left on the table

`j1 + j2` on 2.46M trijets still costs 19.8 ms, and **14.4 ms of that is awkward's gather
floor**: eight `column[index]` fancy-index reads in plain numpy over the same buffers take
14.42 ms, and adding the four sums brings it to 15.82 ms. This PR removes 7.2 ms of the
12.5 ms that sat above that floor.

One awkward-side lever is deliberately not pulled here. `IndexedArray.project` always runs
`awkward_IndexedArray_getitem_nextcarry` to build `nextcarry` from `index`, and for a plain
(non-option) `IndexedArray` the result is byte-identical to `index` — a full copy of a
2.46M-element `int64` buffer, measured at 1.17 ms per operand, so 2.34 ms of the 19.8 ms
above. Skipping that copy belongs in awkward, not in vector, and is not attempted here.

## Relation to #554

This builds directly on #554 ("perf: reduce python overhead for awkward backend", 78cd02f),
which introduced `awkward_transform` and the `_wrap_dispatched_function` hook this PR
extends. #554 fused the *compute* into one layout traversal and reported the same trijet
benchmark going ~25 ms -> 19.1 ms; it left the *reads* alone, so every coordinate field
still cost its own index projection. That is the remaining half, and it is what this
removes. I did not find an issue asking for it — #554's own thread does not mention
carries or projections.

## Tests

`tests/backends/test_awkward.py`:

- an 80-case bit-identity matrix (2 azimuthal x 10 longitudinal/temporal systems x
  generic/momentum names x with/without an option-typed sibling field) comparing `u + v`
  and `u.rho2` against the same computation with `_gather_coordinates` disabled — exact
  `to_list()` and type-string equality;
- extra (non-coordinate) fields survive one-argument ops unchanged;
- index-kernel count: 2 with the gather, 8 without;
- **needed-field witness**: each field holds a distinct constant, so the buffer every
  `NumpyArray._carry` reads names itself — `rho2`/`phi` carry exactly `x, y`, `deltaR`
  exactly `x, y, z` per operand, `tau` exactly `x, y, z, t`. Widening the gather to all
  three groups fails 4 of the 6 cases, so the assertion discriminates;
- nothing is retained on the array after `u+v`, `u.rho2`, `u.deltaR(v)`, and the
  thread-local record is empty afterwards; a dispatch that raises before its call leaves
  no record once the next dispatch wraps;
- unshared indexes are left untouched, record parameters survive the hoisted index, and a
  `RecordArray` shorter than its fields projects only its own rows.

Full suite: `942 passed, 16 skipped` (`pytest -q --ignore=tests/test_notebooks.py`;
skips are ROOT/optree/jax not installed). `pre-commit run --files` on both changed files:
all hooks pass, including `mypy` and `ruff`.

dask-awkward is unaffected: `dak.report_necessary_columns` on `(pairs.u + pairs.v).mass`
still reports `{pt, eta, phi, mass}` and on `pairs.u.pt` still reports `{pt}` — `btag` is
never touched, and the computed result matches the eager one exactly.

### DIFF
--- a/src/vector/backends/awkward.py
+++ b/src/vector/backends/awkward.py
@@ -35,6 +35,7 @@ from __future__ import annotations
 import functools
 import numbers
 import operator
+import threading
 import types
 import typing
 
@@ -624,6 +625,139 @@ _coordinate_fields_spatial = _azimuthal_fields | _longitudinal_fields
 _coordinate_fields_all = _azimuthal_fields | _longitudinal_fields | _temporal_fields
 
 
+_coordinate_field_groups = {
+    "azimuthal": _azimuthal_fields,
+    "longitudinal": _longitudinal_fields,
+    "temporal": _temporal_fields,
+}
+
+
+def _shared_index(record: typing.Any) -> typing.Any:
+    """
+    The one ``Index`` behind every field of ``record``, or None if the fields do
+    not all sit behind the same plain ``IndexedArray``.
+
+    ``RecordArray {f: Indexed(a), g: Indexed(b)}`` is what awkward's broadcasting
+    leaves behind after pushing an index inside a record; it is equivalent to
+    ``Indexed(RecordArray {f: a, g: b})``, which materializes with one index
+    kernel instead of one per field.
+    """
+    contents = record.contents
+    if not contents or record.is_tuple:
+        return None
+    first = contents[0]
+    if type(first) is not ak.contents.IndexedArray:
+        return None
+    for content in contents[1:]:
+        if type(content) is not ak.contents.IndexedArray or content.index is not (
+            first.index
+        ):
+            return None
+    return first.index
+
+
+def _can_one_carry(layout: typing.Any) -> bool:
+    """
+    Whether ``layout`` holds a record behind one shared index, i.e. whether a
+    one-carry gather has anything to collapse; if not, gathering only adds work.
+
+    Option types are excluded: ``IndexedOptionArray.project`` has to build a mask
+    as well, and the shapes it appears in are not the ones this optimizes.
+    """
+    if layout.is_option or layout.is_union or layout.is_numpy or layout.is_unknown:
+        return False
+    if layout.is_indexed:
+        return bool(layout.content.is_record)
+    if layout.is_record:
+        return _shared_index(layout) is not None
+    return _can_one_carry(layout.content)
+
+
+def _project_one_carry(layout: typing.Any) -> typing.Any:
+    """
+    Materialize the record inside ``layout`` with a single index lookup, or None
+    if it is not one of the shapes :func:`_can_one_carry` accepts.
+    """
+    if layout.is_option or layout.is_union or layout.is_numpy or layout.is_unknown:
+        return None
+    if layout.is_indexed:
+        return layout.project() if layout.content.is_record else None
+    if layout.is_record:
+        index = _shared_index(layout)
+        if index is None:
+            return None
+        # a RecordArray may be shorter than its fields; project only its own rows
+        return ak.contents.IndexedArray(
+            index[: layout.length],
+            ak.contents.RecordArray(
+                [content.content for content in layout.contents],
+                layout.fields,
+                parameters=layout._parameters,
+            ),
+        ).project()
+    content = _project_one_carry(layout.content)
+    return None if content is None else layout.copy(content=content)
+
+
+def _gather_coordinates(array: typing.Any, groups: typing.Iterable[str]) -> typing.Any:
+    """
+    Materialize exactly the coordinate fields that ``groups`` names, with a single
+    index lookup, or return None when there is nothing to gain.
+
+    Awkward projects an ``IndexedArray`` once per field asked of it, and every
+    projection re-runs the ``getitem_nextcarry`` kernel over the whole index. An
+    operation reading ``k`` fields of an indexed record therefore pays ``k`` index
+    kernels and ``k`` gathers; done once for the fields the dispatched function
+    actually reads it costs one kernel and the same ``k`` gathers. That index
+    kernel is most of the cost of arithmetic on ``ak.combinations``/
+    ``ak.cartesian`` output.
+    """
+    if not isinstance(array, ak.Array):
+        return None
+    layout = array.layout
+    # A typetracer has no data to gather, and gathering would touch buffers the
+    # calculation may not need (dask-awkward reads those touches to project columns).
+    if not layout.backend.nplike.known_data or not _can_one_carry(layout):
+        return None
+    wanted = frozenset().union(*(_coordinate_field_groups[group] for group in groups))
+    names = [name for name in layout.fields if name in wanted]
+    if len(names) < 2:
+        return None
+    # slicing first carries ``behavior`` (and named axes) over without rebuilding
+    # them, and drops the fields this operation does not read
+    selected = array[names]
+    projected = _project_one_carry(selected.layout)
+    if projected is None:
+        return None
+    selected.layout = projected
+    return selected
+
+
+class _Operand:
+    """One vector's coordinate objects for the duration of one dispatch."""
+
+    __slots__ = ("array", "coordinates", "groups", "source", "wrapped")
+
+    def __init__(self, array: typing.Any) -> None:
+        self.array = array
+        self.source = array
+        self.groups: set[str] = set()
+        self.coordinates: dict[str, typing.Any] = {}
+        # set by a wrap, cleared by the next request: a record still wrapped at
+        # the following wrap belongs to a dispatch that raised before its call
+        self.wrapped = False
+
+
+_dispatch_local = threading.local()
+
+
+def _dispatch_operands() -> dict[int, _Operand]:
+    operands: dict[int, _Operand] | None = getattr(_dispatch_local, "operands", None)
+    if operands is None:
+        operands = _dispatch_local.operands = {}
+    return operands
+
+
 def _yes_record(
     x: ak.Array,
 ) -> float | ak.Record | None:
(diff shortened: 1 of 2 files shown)


henryiii:

Pushed a simplification of the mechanism. The field arrays that one dispatch extracts from an `ak.combinations` operand all share the same `Index` object, and that identity is still visible in `awkward_transform.__call__`. So the one-kernel gather now groups the awkward arguments by index identity right before `ak.transform`, in one helper (`_gather_shared_indexes`).

Removed: the thread-local operand table, the `wrapped` cleanup protocol, the `_coordinates` hook in the twelve coordinate properties, and the layout walkers that hoisted a shared index out of a record. The gather no longer depends on the call order inside `dispatch` and keeps no state between calls. Net -205 lines against `main`.

Timing on 4.1M jet triplets is unchanged from the previous commit (about 10-15% faster than `main` for `j1+j2`, `.mass`, `rho2`, `deltaR`). The fixed per-call cost on small indexed arrays is now a few microseconds instead of about 55 us per operand. Flat arrays are unaffected.

lgray:

Thanks — the stateless shape is the right one upstream. For the record, the same cold benchmark (fresh operands per timed iteration, min-of-20, `awkward 2.13.0`, CPython 3.13, macOS arm64; 7a40b06 is the min of two passes) against `main` and the previous revision:

| case | `main` (ms) | 24e6f08 | 7a40b06 | 7a40b06 vs `main` |
|---|---:|---:|---:|---:|
| trijet xyzt `j1+j2` | 27.85 | 19.92 | 19.83 | −28.8 % |
| trijet xyzt `j1+j2+j3` | 41.98 | 31.05 | 31.21 | −25.6 % |
| trijet xyzt `(j1+j2+j3).mass` | 46.47 | 35.15 | 35.69 | −23.2 % |
| trijet ptetaphim `j1+j2` | 109.29 | 102.29 | 102.41 | −6.3 % |
| trijet xyzt `j1.rho2` | 7.48 | 6.31 | 6.34 | −15.3 % |
| trijet xyzt `j1.pt` | 7.86 | 6.69 | 6.74 | −14.2 % |
| trijet xyzt `j1.deltaR(j2)` | 83.55 | 78.81 | 79.01 | −5.4 % |
| option-typed pairs `j1+j2` | 7.41 | 7.06 | 7.18 | −3.2 % |
| flat 100k xyzt `v+v` | 0.99 | 0.66 | 1.00 | +0.5 % |
| flat 100k xyzt `v.mass` | 0.69 | 0.48 | 0.69 | +0.3 % |
| event MET `+` jagged jets (2D) | 3.19 | 2.99 | 3.09 | −3.4 % |

Every indexed row is identical to the previous revision within noise, so the gain this PR was opened for is intact. The flat rows return to `main`: the ~0.3 ms/100k on those came from the previous revision building each coordinate object once per dispatch instead of three times (`hasattr`, `type(...)`, `.elements`), which the simplification drops along with the state. Small in absolute terms; noting it so the trade is visible rather than silent.

lgray:
@henryiii are we good to go here?

lgray:
@henryiii ping

lgray:
@henryiii please approve or continue review.

henryiii (approved):
Yes, you know me, I need lots of pings. :)

PR #742, opened 2026-09-12
feat(docs): add llms.txt and Markdown page versions

henryiii:

Adds LLM-friendly docs output with NVIDIA's `sphinx-llm` (`sphinx_llm.txt` extension). It runs the markdown builder alongside the HTML build and merges the results, so the whole thing is one extension and one build:

- `llms.txt` — sitemap with a one-line description per page.
- `llms-full.txt` — all docs in one file.
- `<page>.html.md` next to each `<page>.html`.

`llms_txt_description` is set because the default summary is the full README. The index page got a `description` meta tag, since `llms.txt` otherwise used the logo image line as its summary.

Same change as scikit-hep/uhi#252. Built locally with no new warnings.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -42,6 +42,7 @@ extensions = [
     "sphinx.ext.napoleon",
     "sphinx_copybutton",
     "sphinx_math_dollar",
+    "sphinx_llm.txt",
 ]
 
 source_suffix = [".rst", ".md"]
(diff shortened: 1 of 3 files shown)


Saransh-cpp (approved):
Thanks, @henryiii. Looks good! Feel free to merge.

